PR #34, opened 2021-02-15
Mount external checkout directory in /application

maciejzgadzaj:
Ok, here as promised. For you it might be more of a POC rather than finished solution (as it would require building new image for dockerhub), works fine for me though on my RPi, where I need to build my own images anyway.

This essentially mounts external checkout directory in the container, rather than copying all the files to it during the build phase (which requires rebuilding the container when anything changes). So you can edit files, change branches, and whatever only floats your boat outside, and you'll have all the changes reflected in the running container.

The docker command to run it would be something like
```
docker run -it -v <PATH_TO_CHECKOUT>:/application <PATH_TO_CONFIG>:/config:rw meisnate12/plex-meta-manager
```
(not tested, as I'm running it with `docker-compose`)

while my `docker-compose.yml` service definition (as a reminder, this is for RPi, with arm architecture, so I couldn't use your pre-generated image, has to build my own for my archi) right now is this:

```
services:
  plex-meta-manager:
    container_name: plex-meta-manager
    build: Plex-Meta-Manager/
    restart: unless-stopped
    network_mode: host
    volumes:
      - ./Plex-Meta-Manager:/application
      - ${CONFIG_DIR}/plex-meta-manager:/config

```
(repo code is cloned to `Plex-Meta-Manager` subdir).


meisnate12:
ok I think I get it. I'm super new to docker. So this is more for me vs pushing to every user correct? or is it as long as the users dont define the application folder then it all will just work as normal. 

Side note on [reddit](https://old.reddit.com/r/PleX/comments/lffyt0/introducing_plex_meta_manager_a_plex_auto/gmr7ok6/) i was advised to add the command 
`apt-get install -y gcc g++ libxml2-dev libxslt-dev libz-dev && \`
into the dockerfile to make it work for Raspberry Pi 4 (arm) I dont know alot about Docker so do you know whats this does and if it would break other systems to have it in there?

maciejzgadzaj:
It'd be for you, and potentially for those building their own container using the `Dockerfile`. Not for those using the dockerhub image - though I would suspect the same/similar change could be reflected there too (no idea tbh, never built an image for dockerhub before).

Re side note - that user on reddit was me. :) This line generally fixes an error (several of them actually) you'd get when you try to build the container on the RPi:
- `gcc` and `g++` - these are essentially compilers used to build binary libraries from sources - was kinda surprised these were not included in the original `python:3-slim` image, but building was failing without them,
- `libxml2-dev`, `libxslt-dev` and `libz-dev` - are headers (development libs) used to build other libs (dependencies essentially).

Re would it break or not - hard to say really. These will probably work fine for most home users (Debians, Ubuntus etc), while would need small name changes for more rare Linuxes (Red Hats, CentOSes etc). But then again, probably not many people are going to build their own container after all? Well, your decision really how you want to approach it - in either case, probably adding it at least somewhere in the installation info could help others like me.

meisnate12:
lmao I'm dead about referencing your own comment. Thanks for the explanation on that line I made some changes to the dockerfile in dev so we will see how that works. The application folder being able to update is really cool and something I'll remember for the future if I need it but for this project, I just run it locally on my computer while coding I only use the docker to have it run each night and I use docker compose and it can update the image really easily.

datdamnzotz:
> * `gcc` and `g++` - these are essentially compilers used to build binary libraries from sources - was kinda surprised these were not included in the original `python:3-slim` image, but building was failing without them,

Welcome to the world of Docker!

Slim images are the images you want to use *AFTER* you have built your application.

Whole idea here is that you only have what you need in your image binary wise to run your application.  It makes it more secure that way not having tools like gcc or g++ or even ping for that matter.  Also size.

I would suggest using the builder method:

https://towardsdatascience.com/how-to-build-slim-docker-images-fast-ecc246d7f4a7

```
# Here is the build image
FROM python:3.8.0-slim as builder
RUN apt-get update && \
		apt-get upgrade -y --no-install-recommends && \
		apt-get install -y tzdata --no-install-recommends && \
		apt-get install -y gcc g++ libxml2-dev libxslt-dev libz-dev && \
COPY requirements.txt /app/requirements.txt
WORKDIR app
RUN pip install --user -r requirements.txt
COPY . /app

# Here is the production image
FROM python:3.8.0-slim as app
COPY --from=builder /root/.local /root/.local
COPY --from=builder /app/ /app/
WORKDIR app
ENV PATH=/root/.local/bin:$PATH
ENTRYPOINT ["python3", "plex_meta_manager.py"]
```

Issue #26, opened 2021-02-12
Feature Request - Command-line param to process only a specific collection 


maciejzgadzaj:
I have the script set up in a docker container, and it runs at 3 am as planned.

But testing different collection configurations (specifically external lists) I am also running it manually and checking the results. 

Would be great if I could provide an extra command line param to make it reprocess a specific collection only, instead of all of them each time.

And yes, I know I can just leave only the collection config I am testing in my `Movies.yml` file, but still... just an idea. :)

meisnate12:
i think i can come up with something

meisnate12:
ok the develop branch has this feature can you test? run test using `-rt`, `--test`, `--run-test`. Add `test: true` to any metadata or collection object to have them run in test mode. I also turned on Debugging when running test mode so there will be more output.

maciejzgadzaj:
Pulled, rebuilt, can confirm that it works as described (so far at least) - thank you very much!

Not quite what I've originally imagined, was thinking more of sth like `python3 plex_meta_manager.py --run --collection=Comedy,Crime` (without having to additionally modify the `Movies.yml` file for this) - but hey, it's your app after all - as long as it works I'm happy. :)

meisnate12:
I choose this way because like you I run multiple docker containers, one to run the schedule and one with the -r option turned on. So I wanted to be able to spin up another that I can run tests from without having to recreate the container each time. this way I can just run the test container

meisnate12:
also leaving open untill pushed to master

maciejzgadzaj:
> So I wanted to be able to spin up another that I can run tests from without having to recreate the container each time. this way I can just run the test container

Not sure exactly how exactly are you using all your containers, but I might have a solution which you might like for this. Coming to you later tonight, if I have time to polish it locally first.

meisnate12:
I'm excited to see what you've got!

maciejzgadzaj:
https://github.com/meisnate12/Plex-Meta-Manager/pull/34

maciejzgadzaj:
Also had some fun and came up with the above MR... 🤷

maciejzgadzaj:
And then again - we're talking *nixes here only. What about Windows? My guess would be it wouldn't work there either, so it's already not 100% generic. Personally I think I'd add it - considering that it's mostly alpine and buster on dockerhub - and see if anyone complains. :D (If it was my repo obviously.)

Tharic99:
Did this break the --run command by itself?

[...]

PR #35, opened 2021-02-15
#26 command-line param to process only a specific collection(s)

maciejzgadzaj:
Ok, so out of sheer curiosity I had a look at how complicated it would be to achieve what I was thinking about originally, and apparently it was not that complicated at all. 

Bear with me though please, this is the very first time I am doing anything in Python, so there might be some silly things in the code. But then again, from quick tests it seems doing what I wanted it to do, meaning:

- `python3 plex_meta_manager.py` will process all defined collections
- `python3 plex_meta_manager.py --collections='Monty Python'` will process only `Monty Python` collection
- `python3 plex_meta_manager.py --collections=Action,Adventure` will process only `Action` and `Adventure` collections
- `python3 plex_meta_manager.py --collections=missing` (where there is no `missing` collection) will complain about `No collection to update`

Not really asking you here to merge it (although be my guest it you want), just wanted to share, and perhaps even ask for a code review if that's ok?

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -426,7 +426,7 @@ class Config:
 
         util.seperator()
 
-    def update_libraries(self, test):
+    def update_libraries(self, test, requested_collections):
         for library in self.libraries:
             logger.info("")
             util.seperator("{} Library".format(library.name))
@@ -435,14 +435,15 @@ class Config:
             logger.info("")
             util.seperator("{} Library {}Collections".format(library.name, "Test " if test else ""))
             collections = library.collections
-            if collections:
+            collections_to_process = (collections.keys() & util.get_list(requested_collections)) if requested_collections else collections
+            if collections_to_process:
                 logger.info("")
                 util.seperator("Mapping {} Library".format(library.name))
                 logger.info("")
                 movie_map, show_map = self.map_guids(library)
                 logger.info(movie_map)
                 logger.info(show_map)
-                for c in collections:
+                for c in collections_to_process:
                     if test and ("test" not in collections[c] or collections[c]["test"] is not True):
                         continue
                     try:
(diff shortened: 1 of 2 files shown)


modules/config.py
```diff
@@ -420,12 +420,13 @@ def update_libraries(self, test):
             logger.info("")
             util.seperator("{} Library {}Collections".format(library.name, "Test " if test else ""))
             collections = library.collections
-            if collections:
+            collections_to_process = (collections.keys() & requested_collections.split(",")) if requested_collections else collections
```
meisnate12: You can use my util.get_list method to make sure the command line arg will work with "," or ", ".
```python
collections_to_process = (collections.keys() & util.get_list(requested_collections)) if requested_collections else collections
```

meisnate12 (changes requested):
This looks good if resolve the conflicts I'll add this into the release i don't mind giving people more ways to run it.

maciejzgadzaj:
Rebased, conflict resolved, `util.get_list()` used.

doctarrwho:
Thanks @maciejzgadzaj and @meisnate12... Tested in develop and it works as advertised. This is very useful for one-off reprocessing of a collection.

doctarrwho:
Spoke too soon! It was working last night, but I saw it fail this morning for the same collection.

```
[2021-02-17 10:34:30,428] [util.py:586]               [DEBUG]    | Traceback (most recent call last):                                                                 |
                                                                 |   File "/mnt/c/repos/unraid/Plex-Meta-Manager/modules/config.py", line 457, in update_libraries
                                                                 |     collectionless = "plex_collectionless" in collections[c]
                                                                 | TypeError: 'set' object is not subscriptable
                                                                 |
[2021-02-17 10:34:30,429] [config.py:1119]            [ERROR]    | Unknown Error: 'set' object is not subscriptable
````

meisnate12:
ok update and try again

doctarrwho:
Fast as always! It works now 👍

PR #102, opened 2021-03-11
Letterboxd: skip miniseries and TV Shows

miguelamaral:
Letterboxd have limited support for TV, which means the module will choke if a list contains a miniseries or TV Show, as they have empty `data-tmdb-id` attributes:

```html
<body data-type="film" data-tmdb-type="movie" data-tmdb-id="">
```

See [Small Axe](https://letterboxd.com/film/small-axe/) or [Wandavision](https://letterboxd.com/film/wandavision/). You can quickly test this by importing this [list](https://letterboxd.com/randomfunnyname/list/small-axe/):

```
Processing Letterboxd List: https://letterboxd.com/randomfunnyname/list/small-axe
Traceback (most recent call last):
  File "[...]/.apps/plex-meta-manager/modules/config.py", line 463, in update_libraries
    builder.run_methods(collection_obj, collection_name, rating_key_map, movie_map, show_map)
  File "[...]/.apps/plex-meta-manager/modules/builder.py", line 744, in run_methods
    elif "letterboxd" in method:                        items_found += check_map(self.config.Letterboxd.get_items(method, value, self.library.Plex.language))
  File "[...]/.apps/plex-meta-manager/modules/letterboxd.py", line 52, in get_items
    movie_ids.append(self.get_tmdb_from_slug(slug, language))
  File "[...]/.apps/plex-meta-manager/modules/letterboxd.py", line 30, in get_tmdb_from_slug
    return self.get_tmdb(f"{self.url}{slug}", language)
  File "[...]/.apps/plex-meta-manager/modules/letterboxd.py", line 36, in get_tmdb
    return int(ids[0])
ValueError: invalid literal for int() with base 10: ''

Unknown Error: invalid literal for int() with base 10: ''
```
Also fixes a typo — `header` argument in `send_request` — and calls `get_tmdb_from_slug` instead of `get_tmdb`, which prevented the module from running at all in the first place.

Bear in mind my python knowledge is next to nonexistent, so there's probably a more robust way to address this.

### DIFF
--- a/modules/letterboxd.py
+++ b/modules/letterboxd.py
@@ -12,7 +12,7 @@ class LetterboxdAPI:
 
     @retry(stop_max_attempt_number=6, wait_fixed=10000)
     def send_request(self, url, language):
-        return html.fromstring(requests.get(url, header={"Accept-Language": language, "User-Agent": "Mozilla/5.0 x64"}).content)
+        return html.fromstring(requests.get(url, headers={"Accept-Language": language, "User-Agent": "Mozilla/5.0 x64"}).content)
 
     def get_list_description(self, list_url, language):
         descriptions = self.send_request(list_url, language).xpath("//meta[@property='og:description']/@content")
@@ -32,6 +32,8 @@ class LetterboxdAPI:
     def get_tmdb(self, letterboxd_url, language):
         response = self.send_request(letterboxd_url, language)
         ids = response.xpath("//body/@data-tmdb-id")
+        if not ids[0]:
+            raise Failed(f"Letterboxd Error: Miniseries / TV Show found at {letterboxd_url}")
         if len(ids) > 0:
             return int(ids[0])
         raise Failed(f"Letterboxd Error: TMDb ID not found at {letterboxd_url}")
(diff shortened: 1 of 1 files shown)


meisnate12:
i mean you did a damn good job at figuring all that out I actually noticed your fork and already submitted the changes a little different to dev. There just wasn't a need for different error messages. but I appreciate you catching these bugs!

miguelamaral:
Ah, that's great! I switched from PAC and I've been enjoying this immensely, thanks for all the work!

Also, are there plans to cache Letterboxd data? Caching the Letterboxd film id [1] as a GUID once it's matched with TMDb, so the importer doesn't need to go through every film page every time it's run.

I've looked at https://github.com/meisnate12/Plex-Meta-Manager/blob/master/modules/cache.py but I doubt I could get that working on my own.

[1] Films have the `data-film-id` attribute in list pages as well so it should be relatively simple. Slugs could work too, but they can change.

meisnate12:
Yea I could get that added to the cache. I would much enjoy not visiting every url of every movie on the list lol

This actually gives me some ideas for other parts of the program to go in the cache.

I'll see what I can get to this weekend.

meisnate12:
ok added the cache to develop and a better checking for TV Shows

miguelamaral:
Seems to be working! Had to update `Letterboxd = LetterboxdAPI(self.Cache)` though.

https://github.com/meisnate12/Plex-Meta-Manager/blob/develop/modules/config.py#L232

Is this right?

meisnate12:
i ended up doing it slightly differently for consistency but yea you were right

miguelamaral:
Running smoothly, thanks again!

There's a typo: `if self.config.Cache:` https://github.com/meisnate12/Plex-Meta-Manager/blob/a919a1bb5919973b7d15f641c57907ead98e6aec/modules/letterboxd.py#L68

Unrelated question, would it be possible to filter using (substrings of) filenames? I'm not sure the API has access to that data?

meisnate12:
Yea I think so the api has access to the file name of the movie

PR #320, opened 2021-07-05
Fix Sonarr.edit_tags not using translation for apply_tags

handsomematt:
apply_tags wasn't being translated properly in Sonarr/Radarr edit_tags method, it would pass `sync` instead of `add` to arr api resulting in invalid behavior.


meisnate12:
Nice catch! I went ahead and fixed this in develop already but for future reference Merge Request should be to develop not master

PR #462, opened 2021-11-27
Check for t_id value before parsing as list

### LINKED ISSUE #461: Bug: imdb_id support produces error - invalid literal for int() with base 10: ''
**Describe the Bug**
Using the `develop` branch (at commit  62532535c9d6d6b2d686716614c36e05c78efc0e), if library caches items with an IMDb id, but no TMDb id, subsequent runs fail with the following error:

```
invalid literal for int() with base 10: ''
```

This issue seems to stem from the changes in the mentioned commit here: [modules/convert.py](https://github.com/meisnate12/Plex-Meta-Manager/commit/62532535c9d6d6b2d686716614c36e05c78efc0e#diff-43908045658bbae01f921945697fa66359708eef9c5efc9fdd4a690fa9c878d2)

This allows new library items to be added to the cache that have an IMDb id, but not a TMDb number. Then when the program tries to read back the cache on a later run, it tries to read out that empty TMDb entry as an Int starting here in [modules/cache.py](https://github.com/meisnate12/Plex-Meta-Manager/blob/develop/modules/cache.py#L115).

**Relevant Collection Config**
- Having a problem with a collection include the collection config from your metadata file.

Not collection specific

**Plex Meta Manager Info**
- Version Number (can be found at the beginning of your meta.log file.
1.13.0-develop1124 

develop branch commit 62532535c9d6d6b2d686716614c36e05c78efc0e as mentioned above

**Link to logs (required)**
- If having an error with a specific collection include the collection.log otherwise please include the full meta.log file on [Gist](http://gist.github.com). _Do not upload attachments_.

https://gist.github.com/snown/09c35bf2bf79097252acae75900c4231

snown:
## Description

Add check on `t_id` while reading cache, in case row doesn't have a TMDb ID specified.

### Issues Fixed or Closed

- Fixes #461 

## Type of Change

## Checklist

### DIFF
--- a/modules/cache.py
+++ b/modules/cache.py
@@ -112,7 +112,8 @@ class Cache:
                 row = cursor.fetchone()
                 if row:
                     time_between_insertion = datetime.now() - datetime.strptime(row["expiration_date"], "%Y-%m-%d")
-                    id_to_return = util.get_list(row["t_id"], int_list=True)
+                    if row["t_id"]:
+                        id_to_return = util.get_list(row["t_id"], int_list=True)
                     imdb_id = util.get_list(row["imdb_id"])
                     media_type = row["media_type"]
                     expired = time_between_insertion.days > self.expiration


snown:
On further inspection, I don't know that this solves the issue, as now each run says it's adding the IMDb only entries to the cache. So there may be something else needed to support items without TMDb numbers. My fear (although unconfirmed) it that this will cause the cache to continuously grow after each pass, which would not be great.

Hopefully in any case this helps track down the issue, or is a start in the right direction. Adding the `if` statement at least allows the program to run and function again.

PR #541, opened 2021-12-24
Pass read_only_config value to ConfigFile

MarkMckessock:
## Description

Thanks @meisnate12 for adding this flag. Just fixing the flag value not being passed to the ConfigFile class.

## Type of Change

- [+] Bug fix (non-breaking change which fixes an issue)


meisnate12:
Lmao I set all that up and forgot that line

Issue #607, opened 2022-01-13
Feature Request: Allow "duration" filter to use values lower than a minute


CherryKitten:
**Is your feature request related to a problem? Please describe.**
Quoting what I said on Discord:
"can the "duration" filter be changed to allow seconds instead of minutes?

I want to exclude music tracks under 20 seconds from a collection, to filter out sound effects and jingles in video game soundtracks

The current minimum of a minute would also filter out some actual music tracks"

**Describe the solution you'd like**
Either change "duration" filter to be in seconds, or allow decimals to go lower than 1. (the latter would keep existing configurations intact)

meisnate12:
you can now use decimals

PR #682, opened 2022-02-02
add if conditional to check if float is a duration attribute

rtsdev:
## Description

add `if` conditional to check if float is a duration attribute in the `validate_attribute` function and change the maximum value to `None` if it is

### Issues Fixed or Closed

When using duration as part of a filter, if the value was greater than 10 it would throw an error due to float attributes having a hard-coded maximum. This bug was introduced in PR #652 when implementing feature in issue #607.

## Type of Change

## Checklist


meisnate12:
this should be in develop now

PR #1004, opened 2022-08-21
Add last_episode_aired_or_never filter

axsuul:
## Description

Adds `last_episode_aired_or_never` collection filter, similar to `last_episode_aired` but also won't filter out shows that haven't aired yet (i.e. upcoming shows). Use case is I want to include recently aired shows as well as any upcoming shows (i.e. not yet aired) from Trakt trending

```
'Trending Shows':
    collection_order: custom
    trakt_trending: 125
    sync_mode: sync
    filters:
      last_episode_aired_or_never: 100
```

If I were to run this then it won't include any shows that are having a series premiere in the next 24 hours but haven't aired yet

```
'Trending Shows':
    collection_order: custom
    trakt_trending: 125
    sync_mode: sync
    filters:
      last_episode_aired: 100
```

I'm open to calling this attribute another name.

## Type of Change

## Checklist

### DIFF
--- a/docs/metadata/filters.md
+++ b/docs/metadata/filters.md
@@ -125,13 +125,14 @@ Date filters can **NOT** take multiple values.
 
 ### Attribute
 
-| Date Filters                      | Description                                                     |  Movies  |  Shows  | Seasons  | Episodes | Artists  |  Albums  |  Track   |
-|:----------------------------------|:----------------------------------------------------------------|:--------:|:-------:|:--------:|:--------:|:--------:|:--------:|:--------:|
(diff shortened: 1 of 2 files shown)


YozoraXCII:
I'm not sure I understand the use case here - if the item hasn't aired yet then it won't exist in Plex?

chazlarson:
Shows will somewhat frequently get leaked in advance of their air date.  A while ago, for example, a season of Hell's Kitchen appeared, in its entirety, on Usenet in February or so when the show wasn't scheduled to air until April.  That show was on the top of my "recently added" for the next couple months.

axsuul:
Yep exactly shows will frequently leak a day earlier and so the existing filter would exclude that.

meisnate12:
could you change the PR to go to nightly instead of develop

PR #1118, opened 2022-10-17
amend schedule page to remove implication that overlay files can be individually scheduled

YozoraXCII:


YozoraXCII:
Already covered by Sohjiro.

PR #1270, opened 2023-01-13
Change error text

chazlarson:
## Description

This error is related to reading libraries from the config file; the existing error text implies that Plex is implicated in some way when PMM has not yet tried to connect to Plex.  This leads troubleshooters down a blind alley.

## Type of Change

## Checklist


chazlarson:
Not really a "bug fix" per se, but it's the closest match.

PR #1365, opened 2023-04-06
feat: allow env variable substitution

buroa:
## Description

Allows parsing of the configuration file which can include environmental variable substitutions like so:

```
...
tmdb:
  apitoken: !env_var ${TMDB_API_TOKEN}
...
```

As long as the `TMDB_API_TOKEN` env variable is set, this will set `apitoken` to the appropriate value.

## Type of Change

## Checklist

### DIFF
--- a/modules/util.py
+++ b/modules/util.py
@@ -952,15 +952,14 @@ class YAML:
         self.yaml.indent(mapping=2, sequence=2)
         try:
             if input_data:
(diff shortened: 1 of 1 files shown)


meisnate12:
added [Config Secrets](https://metamanager.wiki/en/nightly/home/environmental.html#config-secrets)

PR #1375, opened 2023-04-11
[1] fix ustv-14 names

bullmoose20:


meisnate12:
ill be honest im not a fan of the naming scheme

PR #1385, opened 2023-04-17
Fix mal_search for "sfw: false"

sw-carlos-cristobal:
## Description
No valid logic for handling when "sfw" was set to false.

This minor change fixes the logic to directly parse the true/false state because the else condition was missing.

image
vs.
image

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

## Checklist


meisnate12:
you dont wanna actually do sfw = false as thats just the default state i adjusted the code to fix the above error

Issue #1496, opened 2023-07-07
IMDB collections not working


ryant523:
### Version Number

1.19.0

### What branch are you on?

master

### Describe the Bug

The IMDB Chart collections are no longer being updated.

I'm not sure when this started, but I traced this to to receiving a 403 forbidden error when retrieving the IMDB resources, https://www.imdb.com/chart/moviemeter.

The User-Agent header is not being overwritten, which is causing IMDB to reject these requests.

**modules/imdb.py**

```python
def _ids_from_chart(self, chart):
    if chart == "box_office":
        url = "chart/boxoffice"
    elif chart == "popular_movies":
        url = "chart/moviemeter"
    elif chart == "popular_shows":
        url = "chart/tvmeter"
    elif chart == "top_movies":
        url = "chart/top"
    elif chart == "top_shows":
        url = "chart/toptv"
    elif chart == "top_english":
        url = "chart/top-english-movies"
    elif chart == "top_indian":
        url = "india/top-rated-indian-movies"
    elif chart == "lowest_rated":
        url = "chart/bottom"
    else:
        raise Failed(f"IMDb Error: chart: {chart} not ")
    #return self.config.get_html(f"https://www.imdb.com/{url}").xpath("//div[@class='wlb_ribbon']/@data-tconst")
    # works when I comment the line above and add this 
    return self.config.get_html(f"https://www.imdb.com/{url}", headers=util.header()).xpath("//div[@class='wlb_ribbon']/@data-tconst")

```

```
[2023-07-07 12:37:31,918] [plex_meta_manager.py:697]  [INFO]     |================================= Running IMDb Popular Collection ==================================|
[2023-07-07 12:37:31,919] [plex_meta_manager.py:711]  [INFO]     |                                                                                                    |
[2023-07-07 12:37:31,919] [plex_meta_manager.py:712]  [INFO]     | Sync Mode: sync                                                                                    |
[2023-07-07 12:37:31,919] [plex_meta_manager.py:715]  [DEBUG]    |                                                                                                    |
[2023-07-07 12:37:31,919] [plex_meta_manager.py:716]  [DEBUG]    | Builder: imdb_chart: popular_movies                                                                |
[2023-07-07 12:37:31,919] [plex_meta_manager.py:717]  [INFO]     |                                                                                                    |
[2023-07-07 12:37:31,920] [imdb.py:231]               [INFO]     | Processing IMDb Chart: Most Popular Movies                                                         |
[...]

PR #1499, opened 2023-07-09
added default headers for http requests in config.py

Glenn332:
There was an issue with http requests going to IMDb, this should fix it ;)

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -1076,10 +1076,13 @@ class ConfigFile:
             logger.stacktrace()
             logger.error(f"Webhooks Error: {e}")
 
-    def get_html(self, url, headers=None, params=None):
+    def default_headers():
+        return {"User-Agent": "Mozilla/5.0 Firefox/102.0"}
+    
+    def get_html(self, url, headers=default_headers(), params=None):
         return html.fromstring(self.get(url, headers=headers, params=params).content)
 
-    def get_json(self, url, json=None, headers=None, params=None):
+    def get_json(self, url, json=None, headers=default_headers(), params=None):
         response = self.get(url, json=json, headers=headers, params=params)
         try:
             return response.json()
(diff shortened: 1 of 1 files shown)


holl595:
Why has this been closed and not merged in?

JohnFawkes:
> Why has this been closed and not merged in?

https://github.com/meisnate12/Plex-Meta-Manager/commit/b37d1ebb0458f9e1ef5d99b3aac9487ceab5cb0e

Sohjiro added this ^ to fix imdb

Glenn332:
> Why has this been closed and not merged in?

The owner created a fix [commit](https://github.com/meisnate12/Plex-Meta-Manager/commit/b37d1ebb0458f9e1ef5d99b3aac9487ceab5cb0e). However this only solves it for the Plex module, my fix was more global. There were mentions of other services also requiring a custom user agent instead of the default the python package requests comes with. Those requests are easily blocked.

holl595:
It looks like it’s not resolved the issue with collections, another issue appears to have been raised here https://github.com/meisnate12/Plex-Meta-Manager/issues/1496

Glenn332:
> It looks like it’s not resolved the issue with collections, another issue appears to have been raised here #1496

This PR was never merged in

holl595:
Yes, closing this instead of merging it probably wasn’t the best choice

YozoraXCII:
> Yes, closing this instead of merging it probably wasn’t the best choice

As mentioned in #1496, IMDb are currently going through and re-structuring their entire website, they chose to roll it out slowly rather than all-in-one. This PR will not fix that, it will be a case of PMM playing catch-up every time IMDb pushes new versions of new pages.

Glenn332:
> Yes, closing this instead of merging it probably wasn’t the best choice

Please keep in mind that only charts break when scraping them now, do normal lists still work for you?

PR #1528, opened 2023-07-22
Extract file name; re.match only checks the start

### LINKED ISSUE #1457: dimensional_asset_rename renaming formatted assets
### Version Number

1.19.0

### What branch are you on?

master

### Describe the Bug

When dimensional_asset_rename is true PMM will rename any file in the asset folder even if it is already formatted to another purpose.

Example:
 
`assets > Obi-Wan Kenobi > Season01.jpg` will be renamed to `poster.jpg` even though it has a valid asset name.

Assets with valid names should be ignored with dimensional_asset_rename  is on:

- Season##.ext
- Season##_background.ext
- S##E##.ext
- S##E##_background.ext

This allows me/user to keep the Plex assigned poster but override season posters or episode posters, while also not having to rename posters.ext when adding to asset directory.

This can be solved by just turning dimensional_asset_rename off and taking a few seconds to rename the file

```yml
libraries:                                 
  TV Shows:
    ...
    operations:
      assets_for_all: true
settings:
  dimensional_asset_rename: true
```

### Logs

_No response_

chazlarson:
## Description

re.match() only checks the beginning of the string, so something like:
```
'config/assets/3 1 2 Hours (2021) {imdb-tt13475394} {tmdb-847208}/Season01_background.jpg'
```
Wouldn't get stopped by the dimensional-asset-rename gate and would get renamed.

Now we grab just the filename [`Season01_background.jpg`] and do the re.match() on it.

I did this rather than use re.search() to guard against something like:
```
config/assets/That Movie About S01E04 of Friends (2021)/this-should-get-renamed.jpg
```

### Issues Fixed or Closed

- Fixes #1457

## Type of Change

## Checklist

### DIFF
--- a/modules/plex.py
+++ b/modules/plex.py
@@ -3,6 +3,7 @@ from datetime import datetime, timedelta
 from modules import builder, util
 from modules.library import Library
 from modules.util import Failed, ImageData
+from pathlib import Path
 from PIL import Image
 from plexapi import utils
 from plexapi.audio import Artist, Track, Album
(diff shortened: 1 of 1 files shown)


meisnate12:
Added in a slightly different way [here](https://github.com/meisnate12/Plex-Meta-Manager/commit/596f9bbbd050d7d4d84a375e7607983b1a9376aa)

PR #1606, opened 2023-09-18
display PlexAPI version in header

chazlarson:
## Description

Adds a line displaying the PlexAPI version in use.  Yes, it's pinned in the requirements, but all three branches are currently different and this might be useful information if the user has manually changed it.

## Type of Change

## Checklist


darthShadow:
May be cleaner to use the pre-defined constants in PlexAPI: https://github.com/pkkid/python-plexapi/blob/master/plexapi/const.py

PR #1661, opened 2023-10-16
Modified video_format to support more formats

tylerreed19:
## Description

CHANGES: Split 'WEB' into 'WEB-DL' and 'WEB-RIP'
Added: 'UHD BLU-RAY' | 'HDRIP' | 'SDRIP' | 'HDTV' | 'SDTV' | 'CAM'

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

Adds more options for video format overlays

## Checklist

videoformats


YozoraXCII:
Is there a reason you've created a new file `all_video_format.yml` instead of editing the existing `video_format.yml`?

Your PR also deletes three existing Overlay files

chazlarson:
Do we actually want to open the door to supporting more than TRaSH naming?
```
#                             Utilized tinyMediaManager naming scheme                        #
```

JohnFawkes:
> Do we actually want to open the door to supporting more than TRaSH naming?
> 
> ```
> #                             Utilized tinyMediaManager naming scheme                        #
> ```

I wouldnt be open to it no. This is why we have things like - file: for people to run these things locally and the config repo if they want to share things like this

tylerreed19:
It is really just a modified version of the default yes. And as to the
tinymediamanager naming scheme that was unimportant, it's a really simple
renamer for media. The main point is whatever naming scheme this works for
people with more versions of sources for their media. Which is why I gave
examples of random media types that would work with this.

chazlarson:
So is your suggestion that this be added to the tool as a separate YAML like you have it here, with all the attendant new docs, or were you thinking that this would become part of the existing file, in which case why didn't you PR that?

We are aware of what tinyMediaManager is.  The point of the file naming thing is that currently, all the defaults are assuming a specific naming format, TRaSH, so if a user names their files that way the defaults *will* work as expected.  

If we have some stuff that expects TRaSH naming and some stuff that expects TMM naming, that's then more work to ensure that all the defaults work with either or both.  Avoiding that ever-expanding pool of work is why we settled on TRaSH naming in the first place.  If there are TRaSH naming standards for these additional sources, then please alter this PR to use that naming standard.

I can't see adopting any new default overlay that relies on filename but doesn't work with the standard we specify as the expected naming standard.

tylerreed19:
I understand. I've been utilizing a customized TMM naming scheme that plays exceptionally well with PMM and have had no issues but I completely understand not wanting to add a whole new naming scheme. I've submitted it instead to the Custom Configs. I'll leave it up to you all if you would like to introduce it. I can also add my custom naming scheme data I use in TMM if that would be helpful. Or just leave it out altogether. (I have removed the "Utilizing tinyMediaManager naming scheme" from the description in my config submission).

It may be for a niche portion of PMM users but personally I know I like to know what the true media source is and the naming isn't confusing anything more in PMM. But that's just my opinion. Anyways thanks for looking.

PR #1718, opened 2023-11-08
[Draft] Add Optional Score Filtering to AniList UserList Builder

ZhunCn:
## Description

Add Score Filtering functionality to the AniList Builder for UserList Queries, which allows setting a specific `score_format` for the query to return, a `score` value used for comparison, and a `modifier` field. AniList allows different score formats to be used on a per account basis, so this will allow the query to transform the score based on the user's choice.

Since the GraphQL query does not support score filtering by default, I appended the functionality within the `_userlist` function within `anilist.py`. Because I have done `None` checks for specific arguments, this should be a non-breaking change and will not affect current configurations.

Main reason for adding this feature is so I can auto filter based on my own Completed List and only put shows that have been rated higher than a certain score within the Collection. Because this was a quick code change done late at night, I haven't tested it and have marked the PR as a Draft. I will continue to work on this to fully test the functionality and document the changes. I did also want to get feedback on whether this feature would be in the scope of the project.

## Type of Change

## Checklist

- [ ] I have tested the code to ensure functionality with no regression.
- [ ] I have documented the functionality on the corresponding/relevant files.

### DIFF
--- a/modules/anilist.py
+++ b/modules/anilist.py
@@ -36,7 +36,7 @@ media_source = {
     "original": "ORIGINAL", "manga": "MANGA", "light_novel": "LIGHT_NOVEL", "visual_novel": "VISUAL_NOVEL",
     "video_game": "VIDEO_GAME", "other": "OTHER", "novel": "NOVEL", "doujinshi": "DOUJINSHI", "anime": "ANIME"
 }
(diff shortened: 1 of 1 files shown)


YozoraXCII:
I notice there are no updates to the wiki pages, I'm guessing any new functionality the user can choose from should be detailed in the wiki

meisnate12:
@ZhunCn ive implemented it in a different way see the nightly wiki for how to use can you test it for me?

https://metamanager.wiki/en/nightly/builders/anilist/#anilist-userlist

PR #1823, opened 2024-01-13
Radarr enhancement and new builder

hjone72:
## Description

These three separate commits could be their own pull request because I think they're all valid.

# 1 New builder: JsonUrl
I had a pretty specific use case here. I wanted to be able to use Tautulli's non api endpoint to generate a list of movies from the media_info endpoint. The non api version is a little more powerful and lets you set multiple sort orders. This JsonUrl module will let you process a list of rating keys from a URL.
The module is working but the checking it does isn't as advanced as the checking performed by the Tautulli module. A further enhancement would be to use the same method that Tautulli does to check each rating key.

# 2 Radarr Monitor existing movies
This is a small change that allows you to re-monitor a movie in Radarr that was previously unmonitored.
The use case here is that you have something in radarr that has now been unmonitored. The item now appears on one of the popular lists. In this circumstance I want to monitor it again and get it next time it appears in the RSS. This option basically copies how 'Upgrade existing' works but sets the 'Monitor' back to true for that item.

# 3 Upgrade existing added to the base template
I was scratching my head for longer than I'd like to admit about why this wasn't working. I found these options were missing from the base template. Not sure if it was known and intentional, but this commit allows you to 'Upgrade existing' at an individual collection level.

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

## Checklist


meisnate12:
can you split the builder and the arr change into separate PRs

hjone72:
Sure thing! And the template stuff? I've just noticed that "ignore_cache" is left out too.

Because if cache is enabled the upgrade components are skipped. Not sure if an exception or a comparison should be used, or if just ignoring cache for these known situations.

Thanks,

PR #1825, opened 2024-01-14
JsonUrl builder

hjone72:
## Description

# New builder: JsonUrl
I had a pretty specific use case here. I wanted to be able to use Tautulli's non api endpoint to generate a list of movies from the media_info endpoint. The non api version is a little more powerful and lets you set multiple sort orders. This JsonUrl module will let you process a list of rating keys from a URL.
The module is working but the checking it does isn't as advanced as the checking performed by the Tautulli module. A further enhancement would be to use the same method that Tautulli does to check each rating key.

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

## Checklist

### DIFF
--- a/modules/builder.py
+++ b/modules/builder.py
@@ -1,7 +1,7 @@
 import os, re, time
 from arrapi import ArrException
 from datetime import datetime
-from modules import anidb, anilist, flixpatrol, icheckmovies, imdb, letterboxd, mal, plex, radarr, reciperr, sonarr, tautulli, tmdb, trakt, tvdb, mdblist, util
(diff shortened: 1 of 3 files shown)


meisnate12:
could you explain how to get to this jsonurl?

hjone72:
The idea is more around flexibility. I'm using a custom query against Tautulli to generate a json file which I'm serving via a web server. As long as the JSON contains the rating key it should work.

chazlarson:
This should probably have some documentation added.

YozoraXCII:
@hjone72 do you plan on addressing the documentation suggestion? if not, can we close this?

YozoraXCII:
Closing this as no response from author.

PR #1952, opened 2024-03-27
Small regex fix for DV HDR (resolution overlay), after DV HDR and DV HDR10+ got split

jz1:
## Description

Fix for:
Overlay Error: Regular Expression Invalid: (?i)\bdv((.hdr10?\b)

Removed the extra **(** that was left over of the old OR group with DV HDR10+

### Issues Fixed or Closed

- Related to PR #1947

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] My code was submitted to the nightly branch of the repository.

### DIFF
--- a/defaults/overlays/resolution.yml
+++ b/defaults/overlays/resolution.yml
@@ -217,7 +217,7 @@ templates:
           - alt: plus
             value: '(?i)\bhdr10(\+|p(lus)?\b)'
           - alt: dvhdr
-            value: '(?i)\bdv((.hdr10?\b)'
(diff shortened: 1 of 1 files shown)


adam-rudd:
Does this only work on filenames or metadata?  
I'm seeing over here: https://github.com/meisnate12/Plex-Meta-Manager/discussions/1987
HDR detection works regardless of filename.  

If intended function is to work like HDR off of metadata, then HDR10+ might have bug.

PR #1964, opened 2024-04-04
use radarr monitor values in config.yml

ymendel:
## Description

In a definition, `radarr_monitor` can be movie/collection/none, but in the config.yml where radarr is defined, monitor has to be a boolean. Sonarr can take the full values in both places.
    
This just copies the Sonarr code over so Radarr uses the defined monitor values instead of insisting on a boolean. However, it will turn configs with `monitor: false` into monitoring movie. Not sure what to do about that.

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

Not sure how much this is "bug fix" or "breaking change", honestly.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation change (non-code changes affecting only the wiki)

## Checklist

### DIFF
--- a/docs/config/radarr.md
+++ b/docs/config/radarr.md
@@ -29,7 +29,7 @@ radarr:
   upgrade_existing: false
   monitor_existing: false
   root_folder_path: S:/Movies
-  monitor: false
+  monitor: none
   availability: announced
   quality_profile: HD-1080p
   tag: pmm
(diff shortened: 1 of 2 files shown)


meisnate12:
i had made this change before and it presented issues with older radarr version so it had to be rolled back ill take another look at this to see if it can be added now

ymendel:
Yeah, someone (Yozora?) mentioned that on Discord. And I get it if we need to stay at true/false. But maybe this can be a combo of the values and boolean? It's not a huge deal if this can't be done.

PR #2001, opened 2024-04-24
Fix for non_item_remove_label to always run when defined

jz1:
## Description

It need's to run even when there's no collection items, to remove the labels as defined in the description of the function:

> Matches every movie/show that has the given label and is not in the collection and removes the label

Before it was not removing labels if it was on a collection with 0 matches. It should then still remove the labels defined in non_item_remove_label. 

How to reproduce:
Add a label 'test' to a item.
Then run a collection that has 0 matches, with non_item_remove_label: test
(For example it could happen when TMDB data gets changed and suddenly it doesnt find the correct language for it anymore, it should remove the label even when it has 0 matches for it anymore.)

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

## Checklist

### DIFF
--- a/kometa.py
+++ b/kometa.py
@@ -885,6 +885,9 @@ def run_collection(config, library, metadata, requested_collections):
         except NonExisting as e:
             logger.warning(e)
             library.status[str(mapping_name)]["status"] = "Ignored"
+            if builder.item_details:
+                no_items = True
+                builder.update_item_details(no_items)
         except NotScheduled as e:
             logger.info(e)
             if str(e).endswith("and was deleted"):
--- a/modules/builder.py
+++ b/modules/builder.py
@@ -3153,10 +3153,19 @@ class CollectionBuilder:
         if not self.items:
             raise Failed(f"Plex Error: No {self.Type} items found")
 
-    def update_item_details(self):
-        logger.info("")
-        logger.separator(f"Updating Metadata of the Items in {self.name} {self.Type}", space=False, border=False)
-        logger.info("")
+    def update_item_details(self, no_items=False):
+        if (no_items and "non_item_remove_label" in self.item_details) or not no_items:
+            logger.info("")
+            logger.separator(f"Updating Metadata of the Items in {self.name} {self.Type}", space=False, border=False)
+            logger.info("")
+
+        if "non_item_remove_label" in self.item_details:
+            rk_compare = [item.ratingKey for item in self.items]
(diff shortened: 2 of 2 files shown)


meisnate12:
does it work if you add `ignore_blank_results: true` to the definition

jz1:
> does it work if you add `ignore_blank_results: true` to the definition

I've changed to draft, since it needs more work, and the pull request i did seems not the correct way actually thinking about it again. becouse i let it run after the exception, but then it won't run still with ignore_blank_results: true
But not sure if it should always run even with ignore_blank_results true or false, or only when ignore_blank_results: is true?

But ignore_blank_results: true does get past the exception.
Line 822-823 

>                if not builder.found_items and not builder.ignore_blank_results:
>                     raise NonExisting(f"{builder.Type} Warning: No items found")

but gets halted here
line 867-878

>                 try:
>                     builder.load_collection_items()
>                 except Failed:
>                     logger.info("")
>                     logger.separator("No Items Found", space=False, border=False)
>                 else:
>                     if builder.item_details:
>                         builder.update_item_details()
>                     if builder.custom_sort:
>                         builder.sort_collection()
>                     if builder.sync_to_trakt_list:
>                         builder.sync_trakt_list()

becouse it can't load_collection_items since its empty.
So else: would needs to be changed to finally: so it can still pass it.  becouse now after the except it wont run futher, with a empty collection and even when ignore_blank_results: is set.
But not sure if the other builders still need to run also then, with empty collections?
it could become

>             if valid and run_item_details and (builder.item_details or builder.custom_sort or builder.sync_to_trakt_list):
>                 try:
>                     builder.load_collection_items()
>                 except Failed:
>                     logger.info("")
>                     logger.separator("No Items Found", space=False, border=False)
>                 finally:
>                     if builder.item_details:
>                         builder.update_item_details()
>                     if builder.custom_sort:
>                         builder.sort_collection()
>                     if builder.sync_to_trakt_list:
>                         builder.sync_trakt_list()

or

>                 try:
>                     builder.load_collection_items()
>                 except Failed:
>                     logger.info("")
>                     logger.separator("No Items Found", space=False, border=False)
>                 else:
>                     if builder.custom_sort:
>                         builder.sort_collection()
>                     if builder.sync_to_trakt_list:
>                         builder.sync_trakt_list()
>                 finally:
>                     if builder.item_details:
>                         builder.update_item_details()

YozoraXCII:
Closing as stale

PR #2105, opened 2024-06-06
Switched from retrying to tenacity for http request retries 

ravensorb:
## Description

Switched from the retrying library to the tenacity library for http request retry logic (it is a little more actively in terms of development activity)

## Issues Fixed or Closed

- Fixes 429 errors that can occur due to throttling on sites like ThePosterDB

## Type of Change


meisnate12:
im not opposed to this but your branch is significantly behind nightly so rebase your changes on top of the current nightly

meisnate12:
also `retry_if_http_429_error` and `wait_for_retry_after_header` will never trigger as kometa use the requests module which doesnt throw the HTTPError so those have no error to catch and activate

ravensorb:
> im not opposed to this but your branch is significantly behind nightly so rebase your changes on top of the current nightly

Interesting -- I did a merge just before committing this one.  I'll do another later today.

ravensorb:
> also `retry_if_http_429_error` and `wait_for_retry_after_header` will never trigger as kometa use the requests module which doesnt throw the HTTPError so those have no error to catch and activate

Was thinking about that -- we could use the requests model method raise_for_status or I think we could wrap the requests call in a Retrying().call(...) -- any preference?

meisnate12:
so specifically with adding posters to plex. Kometa just passes the URL to plex so kometa never actually performs the get request to that URL.

meisnate12:
ao ive closed this as just the change from retry to tenacity then in a future PR we can work on the HTTP 429 Errors

ravensorb:
> ao ive closed this as just the change from retry to tenacity then in a future PR we can work on the HTTP 429 Errors

I'll rebaseline and look at the changes again.  In my local testing it did result in a substantial drop in 429 errors when using theposterdb.  I think it is because I am also caching locally (which would make sense).

PR #2147, opened 2024-06-27
rework logic around collection deletion

chazlarson:
## Description

The existing code was deleting collections that didn't match all three criteria.  I made it a little wordier with some trace logging and track each criteria separately.

Example logging while debugging this:

```
| 🎆 Independence Day Movies is smaller than None, DELETE                                             |
| 🎆 Independence Day Movies - managed: False - collection_is_managed: True, NO DELETE                |
| 🎆 Independence Day Movies - configured: False - collection_is_configured: False, DELETE            |
| 🎆 Independence Day Movies would NOT have been deleted by new code                                  |
| 🎆 Independence Day Movies would be deleted by old code                                             |
```

Trace logging produced by PR'ed code:
```
| 🎆 Independence Day Movies - less: None vs collection size: 4, DELETE: False                        |
| 🎆 Independence Day Movies - managed: False vs collection managed: True, DELETE: False              |
| 🎆 Independence Day Movies - configured: False vs collection configured: False, DELETE: True        |
```
## Type of Change


chazlarson:
I don't think this is correct:
```
 return any((delete_less, delete_managed, delete_configured))
```
Don't we want all three to match rather than any one?

That's the issue that sent me down this road in the first place.

This config:
```
delete_collections:
  managed: false
  configured: false
```
Deletes and recreates the independence day collection every time Kometa runs today [with the original code and your reorg of this PR] since it doesn't show up as configured, even though it's managed.  `managed: false` should disqualify it from being deleted as I understand it.

My interpretation of
```
delete_collections:
  managed: false
  configured: false
```
is that collections should be deleted only if they are not managed AND not configured.

meisnate12:
yea you're right it should be `all` not `any`

meisnate12 (approved):

PR #2210, opened 2024-08-29
[WIP] Rework Ratings to not use Plex Ratings

YozoraXCII:
## Description

This change allows Ratings to be fetched from the original source during the Overlays run.

This negates the need to use any "Mass Rating" operations.

In addition, Kometa will automatically select the best `ratingX_image` based on what your `ratingX` choice is. This means the user no longer has to specify `ratingX_image` unless they want to override the default options.

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

- New feature (non-breaking change which adds functionality)
- Documentation change (non-code changes affecting only the wiki)

## Checklist

- Updated Documentation to reflect changes
- Updated the CHANGELOG with the changes


YozoraXCII:
For documentation's sake: This cannot be merged yet due to a design issue whereby Kometa doesn't validate the rating returned using this method, so the rating will say it has been applied even if no rating exists.

PR #2262, opened 2024-10-09
Update kitchen_sink_config.yml

bullmoose20:
updated kitchen sink to reflect letterboxd

## Description

updated kitchen sink to reflect letterboxd

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

- [x] Infrastructure change (changes related to the github repo, build process, or the like)

## Checklist

### DIFF
--- a/json-schema/kitchen_sink_config.yml
+++ b/json-schema/kitchen_sink_config.yml
@@ -1,13 +1,15 @@
 # yaml-language-server: $schema=./config-schema.json
-# .------------------------------------------------------------.
-# | ____        _ _                                ____   ___  |
-# || __ ) _   _| | |_ __ ___   ___   ___  ___  ___|___ \ / _ \ |
-# ||  _ \| | | | | | '_ ` _ \ / _ \ / _ \/ __|/ _ \ __) | | | ||
-# || |_) | |_| | | | | | | | | (_) | (_) \__ \  __// __/| |_| ||
-# ||____/ \__,_|_|_|_| |_| |_|\___/ \___/|___/\___|_____|\___/ |
-# '------------------------------------------------------------'
-### Highly recommend using Visual Studio Code with indent-rainbow by oderwat extension and YAML by Red Hat extension. VSC will also leverage the above link to enhance Kometa yml edits.
-### Must be on Kometa Version 2.0.1-nightly1 or greater
+########################################
+#  _  _____  __  __ _____ _____  _     #
+# | |/ / _ \|  \/  | ____|_   _|/ \    #
+# | ' / | | | |\/| |  _|   | | / _ \   #
+# | . \ |_| | |  | | |___  | |/ ___ \  #
+# |_|\_\___/|_|  |_|_____| |_/_/   \_\ #
+#                                      #
+########################################
+
+### We highly recommend using Visual Studio Code with indent-rainbow by oderwat extension and YAML by Red Hat extension. VSC will also leverage the above link to enhance Kometa yml edits.
+### Must be on Kometa Version 2.1.0-build4 or greater
 libraries:
 # .-------------------------------.
 # | __  __            _           |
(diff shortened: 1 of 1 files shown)


chazlarson:
The position of a lot of things are different, which makes review a little more difficult.

json-schema/kitchen_sink_config.yml
```diff
@@ -777,173 +781,285 @@ libraries:
...
+#  ____  _             _ _     _    #
+# |  _ \| | __ _ _   _| (_)___| |_  #
+# | |_) | |/ _` | | | | | / __| __| #
+# |  __/| | (_| | |_| | | \__ \ |_  #
+# |_|   |_|\__,_|\__, |_|_|___/\__| #
+#                |___/              #
+#####################################
 playlist_files:
-- default: playlist
+  default: playlist
```
chazlarson, line 794: missing dash

json-schema/kitchen_sink_config.yml
```diff
@@ -777,173 +781,285 @@ libraries:
...
+#                           #
+#############################
 trakt:
-  client_id: (redacted)
-  client_secret: (redacted)
   authorization:
     access_token: (redacted)
     token_type: Bearer
-    expires_in: 7889238
+    expires_in: '7889238'
```
chazlarson, line 1038: are quoted strings valid here?

chazlarson (changes requested):

bullmoose20:
Kometa put the quotes strings into my config.yml file.

Kometa does not seem to bark at the missing dash and the config-schema.json in vsc does not bark at it either.

I don't know what to say....

________________________________

chazlarson:
I ask about the quotes because I've never had Kometa add them, so maybe there was some change in Kometa.

The missing dash possibly works because perhaps a list with one item doesn't need them.  I'm just going by the docs, which shows the dash in there.   

That line is showing as changed in this PR, as are the added quotes, for no apparent reason, so I'm asking about them.  This PR is in theory only adding two lines, and there are a lot of things moving around and changing in this file.

bullmoose20:
> I ask about the quotes because I've never had Kometa add them, so maybe there was some change in Kometa.
> 
> 
> 
> The missing dash possibly works because perhaps a list with one item doesn't need them.  I'm just going by the docs, which shows the dash in there.   
> 
> 
> 
> That line is showing as changed in this PR, as are the added quotes, for no apparent reason, so I'm asking about them.  This PR is in theory only adding two lines, and there are a lot of things moving around and changing in this file.

Ok... so let me restart to give us background and then we can update things accordingly.

My new kitchen sink config is based on the output of QS. QS, does not implement libraries and overlay section so that's pretty much the same.

So I took the output of QS which includes the playlist section and merged it with my original kitchen sink and ran Kometa against it.

The config.yml file worked so I then took that and pasted the redacted version of the new config over the one in GH.

Indeed, quotes ended up in the new version that were not there but Kometa added.

Indeed the playlist section is missing a dash and indentation is off but the results worked.

I also thought of reviewing what the yaml and our schema would say about this new kitchen sink config and it was fine with everything except the missing letterboxd in two places that I added.

No complaints on the quotes or on the indentation/missing dash.

Taking all of this, and looking at our wiki, it definitely deviates so we should decide on how to best move ahead.

Option 1(Leaving the wiki alone):
QS output of playlist section needs to be fixed
Kitchen sink playlist section needs proper indentation and dash added
Potential changes to schema.json file
We can't control what Kometa decides to quote ''

Option 2(leave QS output of playlist as is)
Wiki might need updates to reflect additional valid option for playlist section with a different indentation and lack of a dash
Kitchen sink config that was PR'd stays as is
Potential changes to schema.json file
We can't control what Kometa decides to quote ''


I think option 1 is least impacting and since QS is not even mvp, we can fix that to output proper playlist section in MVP. Thoughts?

chazlarson:
I think the wiki should stay as it is, since this:
```
playlist_files:
  - default: playlist
```
takes a list, and if there is more than one item in that list the dashes are required.  I don't think there's any reason to complicate matters by noting the one case in which they aren't.

It could be that ruamel's YAML generation can be tweaked to always add that to lists; QS isn't directly creating any YAML.

My personal opinion is that changes to files like this should be minimal, comprising only what's required.  In this case, the only requirement is the addition of the two letterboxd lines, not changing all the headers and reordering things.

I can see replacing these files with the output of QS when QS can generate a complete config, but I don't see the reason to do so today.  That said, this is just a test file.

bullmoose20:
I agree. Can you kill this commit? We can just add the two letterboxd default lines to the file.

I can re-add the two lines but I don't want to mess up with the delete of this commit.

chazlarson:
https://github.com/Kometa-Team/Kometa/pull/2264

PR #2275, opened 2024-10-16
Don't do the less check on smart collections

### LINKED ISSUE #2274: empty smart collections cause exception when evaluating delete if less condition
### Version Number

2.1.0

### What branch are you on?

master

### Describe the Bug

Here, in operations.py:
```python
            less_check = True
            if less_in is not None:
                less_check = col_in.childCount < less_in
```
We're assuming childCount won't be None.  If it is, like an empty smart collection, Kometa throws:
```
| TypeError: '<' not supported between instances of 'NoneType' and 'int'                             |
```
No need for this.  There should be guardrails.

_No response_

### Logs

_No response_

chazlarson:
## Description

Add a setting to control whether `less` applies to smart collections
Default collection count for comparison to 0 if the count is None

### Issues Fixed or Closed

- Fixes #2274 

## Type of Change

## Checklist


YozoraXCII:
I see that `.envrc` is added, is that intentional here?

Also can't merge because of a conflict in changelog

chazlarson:
> I see that `.envrc` is added, is that intentional here?

Yeah, for those who use direnv

> Also can't merge because of a conflict in changelog

Fixed.

PR #2280, opened 2024-10-17
Add show_unfiltered to allow display of things that *do* match a filter

chazlarson:
## Description

Add `show_unfiltered` setting to allow display of the [possibly] few items that make it through a filter rather than the [possibly] thousands that *don't*

## Type of Change

## Checklist


chazlarson:
This PR also gets rid of a bunch of duplication on the Filters page.

YozoraXCII:
Should we make the default behaviour be true for show unfiltered? Or is this not related to the discussion about overlays last week

chazlarson:
> Should we make the default behaviour be true for show unfiltered? Or is this not related to the discussion about overlays last week

It's related to that discussion , but leaving the default as false won't put anything new in people's logs.  Personally I think the default should be:
```
show_filtered: false
show_unfiltered: true
```
But I left it as is in the interest of least-changes.

YozoraXCII:
> > Should we make the default behaviour be true for show unfiltered? Or is this not related to the discussion about overlays last week
> 
> It's related to that discussion , but leaving the default as false won't put anything new in people's logs.  Personally I think the default should be:
> ```
> show_filtered: false
> show_unfiltered: true
> ```
> But I left it as is in the interest of least-changes.

I believe Soh was in agreement about switching to show items that do meet the filter criteria so that logs aren't hammered by things that don't meet it. I would support this change

chazlarson:
Turning it on *will* cause a bunch of new logging for people who have filtering off.  I'm not sure we want that.  Like a 4K overlay in a 4K library; suddenly the log has thousands more lines.

Maybe leave it off to prevent that.  The reason to add it is to get useful information *if desired*

chazlarson:
This change leaves "show filtered items" off, as has always been the default, *but* enables the user to turn on "just show me the things that made it through" if they want.

YozoraXCII:
I have tested this locally and confirm it works as expected.

When `show_filtered: true`:

image

when `show_unfiltered: true`:

image

PR #2287, opened 2024-10-20
Persisted query not found

chazlarson:
## Description

avoids the stack trace and vague error for this specific failure.

## Type of Change


chazlarson:
Gr.  same thing here.

PR #2310, opened 2024-11-05
Change default to webp_lossless

chazlarson:
## Description

Pillow library now writes XMP data to output files, which results in an error when users attempt to use images [perhaps from TPDB] which have way-too-big [there's a 65K limit in the spec] XMP data; this is a test to see if webp_lossless output files have the same issue.

## Type of Change


YozoraXCII:
Closing as this didn't work

PR #2311, opened 2024-11-05
Revert Pillow Requirement

YozoraXCII:
see #2310 for further detail


YozoraXCII:
Once this PR is merged we should be able to uplift the Pillow requirement again:

https://github.com/python-pillow/Pillow/pull/8483

PR #2347, opened 2024-12-04
docs(PlaylistExample): Add enabling default playlist example

MikeSterry:
## Description
- Add an example showing users how to enable a default playlist

## Type of Change

## Checklist

### DIFF
--- a/docs/defaults/playlist.md
+++ b/docs/defaults/playlist.md
@@ -116,7 +116,7 @@ work. Any value not specified will use its default value if it has one if not it
           use_xmen: false #(2)!
     ```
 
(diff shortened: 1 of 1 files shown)


MikeSterry (commented):
- Small formatting fix

YozoraXCII:
Not a fan of this wording:

"The below is an example config.yml extract showing how to add some default playlists."

Could you change it to:

"The below is an example config.yml extract howing how to enable/disable specific playlists using their keys. By default, all keys/playlists are enabled."

MikeSterry:
Update description per comment

PR #2393, opened 2024-12-31
Fail when user puts definitions in the config.yml

chazlarson:
## Description

Fails when the user puts collection or other definitions directly in config.yml, as:

```yaml
libraries:
  Many-Movies:
    collections:
      Remove Streaming Label:
        build_collection: false
        item_label.remove: Apple TV+ Movies
        plex_search:
          all:
            label: Apple TV+ Movies

    collection_files:
      - default: streaming
...
```

This change makes that a critical error:
```
|========================================= Critical Summary =========================================|
|                                                                                                    |
| Count | Message                                                                                    |
|=======|============================================================================================|
|     1 | The 'Many-Movies' library config contains collections definitions. These belong in external YAML files, not in the config.yml. |
```

## Type of Change

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -220,8 +220,25 @@ class ConfigFile:
         replace_attr(self.data, "save_missing", "plex")
         if self.data["libraries"]:
             for library in self.data["libraries"]:
+
                 if not self.data["libraries"][library]:
                     continue
+
+                def fail_on_definition(found_this):
+                    raise Failed(f"The '{library}' library config contains {found_this} definitions. These belong in external YAML files, not in the config.yml.")
+                if "collections" in self.data["libraries"][library]:
+                    fail_on_definition("collections")
+                if "dynamic_collections" in self.data["libraries"][library]:
+                    fail_on_definition("dynamic_collections")
+                if "metadata" in self.data["libraries"][library]:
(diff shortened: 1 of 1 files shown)


YozoraXCII:
This works as expected when the defintion is entered at the library level, but this produces no error:

```
libraries:
  Kometa-Poster-Testing:
    collection_files:
      collections:
        Remove Streaming Label:
          build_collection: false
          item_label.remove: Apple TV+ Movies
          plex_search:
            all:
              label: Apple TV+ Movies
```

If that's the intention then I'm happy for this to be merged.

chazlarson:
> This works as expected when the defintion is entered at the library level, but this produces no error:
> 
> ```
> libraries:
>   Kometa-Poster-Testing:
>     collection_files:
>       collections:
>         Remove Streaming Label:
>           build_collection: false
>           item_label.remove: Apple TV+ Movies
>           plex_search:
>             all:
>               label: Apple TV+ Movies
> ```
> 
> If that's the intention then I'm happy for this to be merged.

It was only intended to address specific cases we've seen in the wild.  I don't think I've ever seen a `collections` inserted under `collection_files`, since usually someone who understands the latter knows that the former doesn't go there.

PR #2406, opened 2025-01-05
More words about sync mode

chazlarson:
## Description

Add an example for what sync mode does

## Type of Change


YozoraXCII:
Closing as merged with #2409, not sure why this PR is failing on a word that the PR doesn't touch but I addressed that in another PR.

PR #2444, opened 2025-01-12
Languages overlay. big text option

GertSchepens:
## Description

Added `text_size` setting to languages overlay to double the country text font size.
Added `hide_text` setting to languages overlay to only show the flags and hide the country text.

Fixed an issue where `use_subtitles` would ignore `flag_alignment: left` (subtitle flags were hard-coded right.) 

Updated changelog and documentation

### Issues Fixed or Closed

n/a

## Type of Change

- [] Bug fix (non-breaking change which fixes an issue)
- [] New feature (non-breaking change which adds functionality)

## Checklist

- [] Updated Documentation to reflect changes
- [] Updated the CHANGELOG with the changes


YozoraXCII:
Tested and these changes seem to all work as intended.

My only concern is that the real-estate used when `text_size: big` means that these may no longer fit in next to other Defaults files. Generally speaking, the Defaults were designed on a grid so that they could mostly co-exist without overlapping each other.

If @bullmoose20 is happy with this change then I am happy as this is more his thing than mine.

Here's an example of how the "big" text size looks compared to the standard size:

image

GertSchepens:
Thanks :)

Maybe a bit of context on why. 
Screenshot_20250113-012215

The current text is too small to comfortably read in the mobile app. The big version is just ok for the vision impaired people in my house.
If it takes up too much space, I can tighten the space between, though there too, I enjoy the readability it offers on smaller screens.

YozoraXCII:
> Maybe a bit of context on why. ![Screenshot_20250113-012215](https://private-user-
> The current text is too small to comfortably read in the mobile app. The big version is just ok for the vision impaired people in my house. If it takes up too much space, I can tighten the space between, though there too, I enjoy the readability it offers on smaller screens.

Thanks for providing the episode screenshot, I didn't think to test that!

Happy with what you've said but would still like bullmoose to confirm he's happy before this gets merged.

YozoraXCII:
@meisnate12 is doing some testing as he thinks this may be possible via template variables without having to add new attributes, holding off merging until confirmation received

meisnate12:
It's not but I'm doing something else to make this easier so still don't merge but it's not just gonna be template variables

meisnate12:
Replaced with #2465

PR #2453, opened 2025-01-19
Append imdb ID to OMDb API query to resolve ongoing OMDb API issue

### LINKED ISSUE #2452: omdb needs apikey to not be last argument and/or to use https
### Version Number

2.1.0

### What branch are you on?

master

### Describe the Bug

Looks like if the apikey is the last argument, omdb will (sometimes?) treat the api as invalid. See discussion here: https://github.com/omdbapi/OMDb-API/issues/311 and https://github.com/omdbapi/OMDb-API/issues/316 

This was biting me. I changed https://github.com/Kometa-Team/Kometa/blob/7ca6acbe2fd4a3326ffea0641a109be65044e4de/modules/omdb.py#L64 in my local copy to 
```
response = self.requests.get(base_url, params={"apikey": self.apikey, "i": imdb_id})
``` 

and that seemed to fix the issue. 

I also tried changing base_url to just be https and that seemed to work, but per the linked issues, it may be inconsistent for others. 

```yml

```

### Logs

_No response_

YozoraXCII:
Fixes an issue where OMDb would intermittently say an apikey is invalid if the apikey is the last parameter passed (see https://github.com/omdbapi/OMDb-API/issues/316)

Resolved by appending the imdb id (`tt0080684`) to the query as suggested in above link and in #2452 

Closes #2452


YozoraXCII:
Seal of approval

image

Issue #2477, opened 2025-01-24
ASSET_NAME Folder Name validation function removing Colins and Hypens


unmasking-crypto:
### Version Number

2.1.0 (Docker: master)

### What branch are you on?

master

### Describe the Bug

I've had an issue with the ASSET_NAME match within Kometa I've got folder's setup that matches the exact name within Plex however the sanatization function in util.py is causing characters such as colin's and hyphens to to be removed from my Folder name so it can't find the assets folder. 

I added in debug to verify this was the issue the debug printing is shown below.
I'm happy to make a pull request however unsure how this should be rectified. 

I

![Image](https://github.com/user-attachments/assets/1705ccce-e041-4590-8725-02a0e63ccc94)

### Relevant Collection/Overlay/Playlist Definition

```yml

```

chazlarson:
This is by design.  Characters which are potentially illegal in file paths are removed to determine the target asset name.

unmasking-crypto:
I'm not trying to get filename validation removed just fixing valid paths within different operating systems. 
I've taken a look at the documentation for the function is_valid_filepath we can pass in a parameter of platform: "auto" to resolve the issue ? 

I've tested it and it works on my end and it fixes the issue. You should be able to see on patch-1 or something on my fork.

YozoraXCII:
Closing as by design

PR #2478, opened 2025-01-24
changed validate filename to use auto operating system detection.

unmasking-crypto:
## Description

Added a change to the function is_valid_filepath function to check that the filepath is compatible based on the operating system. allowing support for a wider range of possible filenames. 

References the issue #2477 
[docs](https://pathvalidate.readthedocs.io/en/latest/pages/reference/function.html#pathvalidate.is_valid_filepath)

- [] Bug fix - Supporting more filename types based on operating system,

### DIFF
--- a/config/config.yml.template
+++ b/config/config.yml.template
@@ -81,6 +81,7 @@ settings:
   show_missing: true
   show_missing_assets: true
   save_report: false
+  operating_system_type: universal  
   tvdb_language: eng
   ignore_ids:
   ignore_imdb_ids:
--- a/json-schema/config-schema.json
+++ b/json-schema/config-schema.json
@@ -782,6 +782,11 @@
                     "description": "Used to show missing items from collections or playlists.\nWhile show_missing is true items missing from collections or playlists will be displayed.",
                     "type": "boolean"
                 },
+                "operating_system_type": {
+                    "description": "Used to set the operating system type defaults to universal.\nSet the operating system type to either 'Linux', 'Windows', 'macOS', 'POSIX', 'universal'.\nUsed for improving/extending the validation of filenames and paths, taking the operating system into account.",
+                    "type": "string",
+                    "enum": ["Linux", "Windows", "macOS", "POSIX", "universal"] 
+                },
                 "show_missing_assets": {
                     "description": "Used to print a message when assets are missing.\nDisplay missing asset warnings for items, collections, and playlists.",
                     "type": "boolean"
--- a/kometa.py
+++ b/kometa.py
@@ -638,6 +638,7 @@ def run_libraries(config):
                 logger.debug(f"Asset Directory: {ad}")
             logger.debug(f"Asset Folders: {library.asset_folders}")
             logger.debug(f"Asset Depth: {library.asset_depth}")
+            logger.debug(f"Operating System Type: {library.operating_system_type}")
             logger.debug(f"Create Asset Folders: {library.create_asset_folders}")
             logger.debug(f"Prioritize Assets: {library.prioritize_assets}")
             logger.debug(f"Dimensional Asset Rename: {library.dimensional_asset_rename}")
@@ -819,9 +820,9 @@ def run_collection(config, library, metadata, requested_collections):
             logger.separator(f"Resuming Collections")
 
         if "name_mapping" in collection_attrs and collection_attrs["name_mapping"]:
-            collection_log_name, output_str = util.validate_filename(collection_attrs["name_mapping"])
+            collection_log_name, output_str = util.validate_filename(collection_attrs["name_mapping"], config.operating_system_type)
         else:
-            collection_log_name, output_str = util.validate_filename(mapping_name)
+            collection_log_name, output_str = util.validate_filename(mapping_name, config.operating_system_type)
         #logger.add_collection_handler(library.mapping_name, collection_log_name)
         library.status[str(mapping_name)] = {"status": "Unchanged", "errors": [], "added": 0, "unchanged": 0, "removed": 0, "radarr": 0, "sonarr": 0}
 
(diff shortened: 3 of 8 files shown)


YozoraXCII:
Without much knowledge on this aspect of Kometa - if people are using asset folders etc, will this create issue where files with `:` characters are now treated as valid and thus a new asset folder will be created?

chazlarson:
> Without much knowledge on this aspect of Kometa - if people are using asset folders etc, will this create issue where files with `:` characters are now treated as valid and thus a new asset folder will be created?

Yes, it would break (some) existing setups.

It also potentially breaks all third-party scripts that read from or add to an asset directory.

I  could see adding this with a toggle to turn it on if someone wants this behavior.

chazlarson:
What's the use case where this is important enough to risk so much breakage?  The only one I can think of is where one wants to use the media directories as asset directories, and one's media folders contain "illegal" characters.

YozoraXCII:
@dextrous0z can you amend this PR so that it's a toggle (perhaps something within the `settings` section of config.yml?) with the default behavior being disabled.

unmasking-crypto:
@YozoraXCII 
Been giving this some thought and yes @chazlarson is right this could potentially break some existing setups my current use case for this is I've currently started using a fscorrupt/Posterizarr setup and I noticed whilst the asset directories existed some where getting skipped by Kometa as the folder name determined after sanitizing is different to the actual Plex name. 
As my plex folders contain a bunch of characters that probably won't be valid on windows. 


I'm thinking the default behavior will be how it currently originally is and adding a config within the "config.yml" that allows users to set the option based on the operating system for example the spec lists the below as valid: 
        
        "Linux"

        "Windows"

        "macOS"

        "auto"

        "POSIX": POSIX-compliant systems (Linux, macOS, etc.)

        "universal"/None
        
However if this is run in Docker I'm guessing that auto will always default to linux as this is built on alpine. So allowing a user to pick is a better option.

chazlarson:
As someone who maintains one of those third-party scripts that works with the asset directory, I'm not crazy about this complication, but it's not my decision.

meisnate12:
im also not crazy about this complication but while its being discussed there also needs to be validation on the input. Use the `test_list` parameter of the `check_for_attribute` method which takes a dictionary where the key is the attribute to validate and the value is a description. see `run_order`, `sync_mode`, or `overlay_artwork_filetype` for reference.

meisnate12:
also please specifiy the attribute `platform` when calling `sanitize_filename` and `is_valid_filename` right now you're passing the `operating_system` variable to `sanitize_filename` as `replacement_text` not `platform`

miaLDN:
> @YozoraXCII Been giving this some thought and yes @chazlarson is right this could potentially break some existing setups my current use case for this is I've currently started using a fscorrupt/Posterizarr setup and I noticed whilst the asset directories existed some where getting skipped by Kometa as the folder name determined after sanitizing is different to the actual Plex name. As my plex folders contain a bunch of characters that probably won't be valid on windows.
> 
> I'm thinking the default behavior will be how it currently originally is and adding a config within the "config.yml" that allows users to set the option based on the operating system for example the spec lists the below as valid:
> 
> ```
>     "Linux"
> 
>     "Windows"
> 
>     "macOS"
> 
>     "auto"
> 
>     "POSIX": POSIX-compliant systems (Linux, macOS, etc.)
> 
>     "universal"/None
> ```
> 
> However if this is run in Docker I'm guessing that auto will always default to linux as this is built on alpine. So allowing a user to pick is a better option.

Python really only supports two OSes:

- `nt` = Windows
- `posix` = Unix

``` py
isWindows = os.name == 'nt'
```

`os.name` will only return one of 'nt' or 'posix' for Python [3.8](https://github.com/python/cpython/blob/3.8/Lib/os.py#L51) ... [3.13](https://github.com/python/cpython/blob/3.13/Lib/os.py).  Python 2.7 will still return 'nt' or 'posix', and some other dead platforms.  [Python 3.0](https://github.com/python/cpython/blob/3.0/Lib/os.py#L41) you will never get, but has 'ce' and 'riscos'

`platform.system()` returns many aliases and the platform library is for detailed system information.

https://docs.python.org/3/library/os.html#os.name

> `os.name`
> The name of the operating system dependent module imported. The following names have currently been registered: 'posix', 'nt', 'java'.

BTW - the Python docs says `java`, but  in the Python source code only `nt` and `posix` are returned!

miaLDN:
> > Without much knowledge on this aspect of Kometa - if people are using asset folders etc, will this create issue where files with `:` characters are now treated as valid and thus a new asset folder will be created?
> 
> Yes, it would break (some) existing setups.
> 
> It also potentially breaks all third-party scripts that read from or add to an asset directory.
> 
> I could see adding this with a toggle to turn it on if someone wants this behavior.

Definitely not keen here on breaking changes which affect the very core of Kometa, the asset folders etc.

@dextrous0z #2477 doesn't contain much information.  What OS and file-system are you using for your setup.  Personally I'd like to see #2477 bug report with more details added so that we can review the breaking-change requested here in #2478

unmasking-crypto:
@planetrocky I'm running Kometa on Unraid Docker, the file system is brfs. I can provide log files if it helps.
 
Between the two different logs files, the only thing you'll see however is what is mentioned in the original issue #2477 
Which is a bunch of missing Asset Folder lines.

I don't mind making changes to the pull request. If you believe the approach is wrong just tell me how'd you like to see this resolved ? 

For example we can just say Kometa will never allow asset folders not following a Universal layout. Then no action is needed and I can just fix my library folders and rebuild plex. 

Or we,
We can go the route of using python methods to determine the operating system running the program, which would IMO be the cleaner route. 

My original concern with not doing this was due to Windows Docker, which WILL most likely cause issues, as the operating system will be deemed as Linux, and in turn use Linux file validation. 

@meisnate12 First time contributing to the code missed that one, added in the latest commit. 

Also will do the documentation if you want just really don't understand how to see an accurate preview of the .md files correctly.

YozoraXCII:
Closing as I don't believe we are implementing this due to the impact it may have on users.

PR #2496, opened 2025-01-31
specifically state that there is no UI

chazlarson:
## Description

Another "I had no idea whtere was no UI" found in the wild. The lack of any "how to open the UI" was apparently taken as "these docs are incomplete" so it should probably be explicitly stated.

## Type of Change

## Checklist


YozoraXCII:
Added into 2487, closing

PR #2551, opened 2025-03-26
Add `tmdb_deathday`

Kevin2kkelly:
## Description

I made an attempt at adding the ability to create "In Memoriam" collections via `tmdb_deathday`, similar to how birthday collections are able to be created via `tmdb_birthday`.

MASSIVE DISCLAIMER: This involves code changes to `builder.py` and I have no python experience, so this may require a more thorough review (no worries if this is not a priority right now).

I essentially took the existing section that I thought was used for `tmdb_birthday` (lines 774-883), duplicated that, and created a second `tmdb_deathday` section after doing a find & replace on "birthday" for lines 885-994.

I ran this against a test library with just a few movies, and both the birthday and deathday collections were created as expected.

Config:
```
libraries:

  Kometa-Deaths:
    collection_files:

    - default: actor
      template_variables:
        collection_section: '001'
        use_separator: false
        radarr_add_missing: false
        style: signature
        name_format: 🎂 Happy Birthday <<key_name>>
        translation_key: birthday
        sort_by: audience_rating.desc
        title_format: <<key_name>>'s Birthday
        data:
          depth: 10
          limit: 200
        tmdb_birthday:
          this_month: true
          before: 0
          after: 0

    - default: actor
      template_variables:
        collection_section: '001'
        use_separator: false
        radarr_add_missing: false
        style: signature
        name_format: Remembering <<key_name>>
        sort_by: audience_rating.desc
        title_format: Remembering <<key_name>>
        data:
          depth: 10
          limit: 200
        tmdb_deathday:
          this_month: false
          before: 60
          after: 60
```

Log:
[meta.log](https://github.com/user-attachments/files/19458922/meta.log)

Please let me know if I'm missing anything or if this requires additional testing. Thanks!

## Type of Change

## Checklist


YozoraXCII:
I'll let someone with a more technical understandng of the code review this

nwithan8:
Welp, building an auto-memoriam tool for Kometa has been on my backlog for three years, glad someone finally actually sat down and did it, thanks @Kevin2kkelly !

nwithan8:
Heads up, running this on a movie library with 30,000 items in it, trying to load the actors for every movie, caused my machine to run out of memory. It's not necessarily a problem with the death day specifically, more-so with the underlying "tmdb_actor" dynamic collections. Unlike other lists where it might only be looking for a set number of actors as provided by, e.g. a trending list from TMDb, this has to first parse all the actors from every item in the library, and then check the death date for each on TMDb, which requires significantly more processing.

YozoraXCII:
I'm working on a more elegant solution that doesn't hammer TMDb or Kometa, nothing to share yet and not sure it'll come to anything but we shall see

PR #2607, opened 2025-04-10
add log file name prepend arg to allow multiple simultaneously kometa runs

reloxx13:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

Add an optional argument `--log-filename-prepend` to allow prepending a custom string to the log file name. This enables running multiple Kometa instances simultaneously - for example, one instance per library - without file access conflicts.

Without this option, all instances try to write to the same default `meta.log`, which leads to file locking issues or write errors, especially on systems where the log file cannot be shared (e.g., Windows). Using a unique log file per instance solves this problem and improves parallel execution and debugging.

**Example usage:**
```bash
python kometa.py --run-libraries "Movies" --log-filename-prepend movies
python kometa.py --run-libraries "Series" --log-filename-prepend series
```

This will generate `movies_meta.log` and `series_meta.log` respectively.

## Related Issues [optional]
- Related Issue -
- Closes -

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ Update tenacity requirement to 9.1.2
 # Important Changes
 
 # New Features
+- Add argument to prepend a string to the logfilename `--log-filename-prepend movies`
 
 # Docs
 
--- a/docs/kometa/environmental.md
+++ b/docs/kometa/environmental.md
@@ -438,6 +438,36 @@ Kometa will load those environment variables when it starts up, and you don't ha
             docker run -it -v "X:\Media\Kometa\config:/config:rw" kometateam/kometa --overlays-only
             ```
 
+??? blank "Log filename prepend&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`-lfp`/`--log-filename-prepend`&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`KOMETA_LOG_FILENAME_PREPEND`<a class="headerlink" href="#log-filename-prepend" title="Permanent link">¶</a>"
+
+    <div id="log-filename-prepend" />Prepend a string to the log filename. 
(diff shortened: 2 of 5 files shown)


reloxx13:
added docs, but unable to see the result/test the docs.

YozoraXCII:
Can you add logfile and logfilename to `.github/.wordlist.txt`

meisnate12:
hey so i dont plan on merging this you shouldn't run these simultaneously your plex db easily could end up locking up. Plus if you really wanted to you can have a separate instance of kometa and just run that. having this as a feature will push people into thinking that its ok and imo its just not a good idea.

reloxx13:
Sad because this should be my or server owners decision and not yours. 
No you cannot simply run mutliple instances because of the locked logfile (unable to open file, cause file is already opend python error on windows). 

You could add this but dont add the hint for mutiple instances. 
You dont use multithreads and the plex db can handle 10 connections.

I never had problems with locking and i spam 8 connections in the night on it for another own script. Even with more u only get a connection refused cause of 10 open connections limitation.

And at the end, frustration for contributers.   
Cya

meisnate12:
You can have a separate copy of the full kometa folder in two separate locations  and then use `--config` to point to the same config file. so go off and do it.

chazlarson:
If those multiple instances have their own config dirs, there is no logfile conflict.

chazlarson:
Another issue with multiple Kometa from the same config directory with the same config file is that they will be accessing the same sqlite Kometa cache file, which will cause problems.

reloxx13:
You all bring theoretical arguments that don't hold up in practical use.

Great, so no double cache is used, thats the usecase for cache.    
The Plex DB is also not locking up, it would lock up on multiple users using plex.

PR #2641, opened 2025-05-01
Add a new GooglyEyes special filter

zachswasey:
Replaces found eyes in faces with a googly eye image.

* Supports rotating the googly eye, default is gravity down
* Supports increasing sizes if the user wants giant or tiny eyes

## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

This change adds a new special filter (like Blur) that adds googly eyes to posters where eyes in human faces are detected. It allows the user to set the size, rotation (randomly), and confidence, but the defaults yield pretty good results.

overlays-example-googlyeyes

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


chazlarson:
New requirements [tensorflow, specifically] won't install in the docker build; it appears this will require a change in the base image.

Changing the base from `python:3-slim` to `python:3-buster` allows tensorflow to install.

zachswasey:
> New requirements [tensorflow, specifically] won't install in the docker build; it appears this will require a change in the base image.
> 
> Changing the base from `python:3-slim` to `python:3-buster` allows tensorflow to install.

Ah, it looks like the latest version of tensorflow (2.19.0) doesn't support Python 3.13 yet, and that's what `python:3-slim` is using. I've been using Python 3.12 and didn't think to do a `docker build`.

Using `python:3.12-slim` should work for the official Docker image (`python:3-buster` hasn't been updated in 2 years). And since Kometa supports up to Python 3.13, I could move the new requirements to a separate requirements-optional.txt, with a runtime check for functionality, and update the docs I've already added to mention installing the extra requirements for those users running local builds. Let me know what you think. I think this feature would be enjoyed by a lot of users, so I'm very interested in the least disruptive way.

chazlarson:
I also had to specifically install `tf-keras` to get it to run not-in-docker:
```
ValueError: You have tensorflow 2.19.0 and this requires tf-keras package. Please run `pip install tf-keras` or downgrade your tensorflow.
```

chazlarson:
Here's one consideration:
```
kometateam/kometa                            googly             f111ecdd9279   6 minutes ago   2.41GB
kometateam/kometa                            latest             073cab07d927   4 weeks ago     302MB
```
This is just changing the base image to `python:3.12-slim`

That's mostly due to tensorflow:
image

seems kind of expensive, but not my call.

chazlarson:
So this is cool, but not something that's worth increasing the image size by 7x, in my opinion.

Could this be done with a separate tensorflow container, failing gracefully if the API endpoint wasn't available?

https://mohitkr777.medium.com/deploying-a-tensorflow-model-with-tensorflow-serving-and-docker-a-step-by-step-guide-using-a89243dbc424

zachswasey:
That's a good compromise, and one I was mulling over as well after you pointed out the size increase may be an issue for some people. I'll experiment with either just wrapping RetinaFace in a simple Flask service, or just serving the model via TF Serving, but since RetinaFace isn't my library the former is probably the most straight-forward approach.

YozoraXCII:
Closing due to no movement

PR #2784, opened 2025-09-04
Handle operations in batches

d3mystified:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

This PR breaks up operation handling into batches (not to be confused with plex update batches that are already done). This is done to increase reliability and performance of the operations (basically by breaking up a large chunk of work into smaller chunks). It also improves memory usage - by only keeping a batch's worth of updates in memory.

In addition, some reverse proxies reject large URLs, which can happen for a large enough number of edits.

For my very large library (10k+ movies) ratings updates would never finish (because something would cause kometa to break - reverse proxy, server instability, host instability, OOM etc). This change makes it way more reliable - I'm actually getting updates done now.

## Related Issues [optional]

- Related Issue #
- Closes #

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

## Sample logs

```
[INFO]     |====================================================================================================|
[INFO]     |                                      Processing Batch 1/1129                                       |
[INFO]     |====================================================================================================|
[INFO]     |                                                                                                    |
[INFO]     | (1/11288) #1 HAPPY FAMILY USA                                                                      |
... snip
[INFO]     | Batch Audience Rating Update (29/35): 5 Episodes updated to 8.4                                    |
[INFO]     | Batch Audience Rating Update (30/35): 3 Episodes updated to 8.5                                    |
[INFO]     | Batch Audience Rating Update (31/35): 2 Episodes updated to 8.6                                    |
[INFO]     | Batch Audience Rating Update (32/35): 3 Episodes updated to 8.7                                    |
[INFO]     | Batch Audience Rating Update (33/35): 2 Episodes updated to 8.8                                    |
[INFO]     | Batch Audience Rating Update (34/35): 3 Episodes updated to 9.0                                    |
[INFO]     | Batch Audience Rating Update (35/35): 1 Episode updated to 9.3                                     |
[INFO]     |                                                                                                    |
[INFO]     |====================================================================================================|
[INFO]     |                                       Done with Batch 1/1129                                       |
[INFO]     |====================================================================================================|
[INFO]     |====================================================================================================|
[INFO]     |                                      Processing Batch 2/1129                                       |
[INFO]     |====================================================================================================|
```

(Ran with `batch_size=10` just to test out)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@ Updated requests requirement to 2.32.5
 
 # New Features
 Adds `|` as a preferred delimiter for run times.
+Library operations now happen in batches of items (size=100) instead of over the entire library all at one.
 
 # Docs
 Fixed an issue where the home page logo and shield images appeared as raw text
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@ Updated requests requirement to 2.32.5
 
 # New Features
 Adds `|` as a preferred delimiter for run times.
-Library operations now happen in batches of items (size=100) instead of over the entire library all at one.
+Adds an optional setting `plex_bulk_edit_batch_size` in operations to control how many items are processed in a single batch.
 
 # Docs
 Fixed an issue where the home page logo and shield images appeared as raw text
--- a/docs/config/operations.md
+++ b/docs/config/operations.md
@@ -946,3 +946,30 @@ Several of these operations perform **mass** updates; these are just that, **mas
             metadata_files:
               file: config/Movie_Backup.yml  # apply this Metadata File to the library
         ```
+
+###### Plex Bulk Edits Batch Size
+
+??? blank "`plex_bulk_edit_batch_size` - Breaks up processing of operations into chunks of this many items.<a class="headerlink" href="#plex-bulk-edit-batch-size" title="Permanent link">¶</a>"
+
+    <div id="plex-bulk-edit-batch-size" />Setting this file tells Kometa to break up operations that require inspecting/modifying the entire library into chunks. Setting this can be particularly helpful for large libraries.
+
+    If no value is set, the entire library is processed in one batch.
+
+    For example, if your Movies library has 1000 items, by default all of them will be processed at once.
+    If `plex_bulk_edit_batch_size=100`, then 100 items will be processed at once.
+
+    <hr style="margin: 0px;">
+
+    **Attribute:** `plex_bulk_edit_batch_size`
+
+    **Accepted Values:** None, or any number.
+
+    ???+ example "Example"
+
+        ```yaml
+        libraries:
+          Movies:
+            operations:
+                plex_bulk_edit_batch_size: 1000
+                mass_audience_rating_update: imdb
+        ```
--- a/modules/config.py
+++ b/modules/config.py
@@ -149,6 +149,7 @@ library_operations = {
     "mass_originally_available_update": mass_available_options, "mass_added_at_update": mass_available_options,
     "mass_collection_mode": "mass_collection_mode", "mass_poster_update": "dict", "mass_background_update": "dict",
     "metadata_backup": "dict", "delete_collections": "dict", "genre_mapper": "dict", "content_rating_mapper": "dict",
+    "plex_bulk_edit_batch_size": "int",
 }
 
 class ConfigFile:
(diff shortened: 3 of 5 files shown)


YozoraXCII:
This appears to work as expected, I would suggest allowing the user to tweak the batch size to suit their needs.

I would also consider upping the default to 500 items since operations failing due to long URLs seems to be extremely rare.

```yaml
libraries:
  Movies:
    operations:
      batch_size: 25
      mass_genre_update: imdb
```

If the `batch_size` attribute isn't found in config.yml, it would use the default from the module.

YozoraXCII:
Just to add, the only reason I suggest increasing the batch size default is that, based on my very limited test of 2 runs, lower batch size seems to increase the runtime (potentially because of the additional logging that it needs to do).

Doing a simple mass rating update took 44s normally, and with a batch size of 3 took 1m 4s, so that may escalate on large libraries even if the batch size is 100.

antwanchild:
And what if we don’t want them in batches at all, other than the one mass batch?

YozoraXCII:
> And what if we don’t want them in batches at all, other than the one mass batch?

Not sure I follow? Kometa already does one big batch, this lets you lower the amount.

If my suggestion is implemented, you could set the batch to 1 but I don't know why you would since you're losing out on speed gains by batching.

antwanchild:
Right. That’s what I’m asking. I would rather not do in smaller batches. This makes it seem to me that smaller batches will be the new thing instead of one batch (the way it is now)

d3mystified:
I understand the concern. I'll make this backwards compatible.

Meaning if the user has:

```
libraries:
  Movies:
    operations:
      batch_size: 25
      mass_genre_update: imdb
```

Then we do batches of 25 items.

If the user has

```
libraries:
  Movies:
    operations:
      mass_genre_update: imdb
```

Then we do 1 batch of all items.

d3mystified:
Updated the PR with changes as discussed, and updated documentation as well.

### No batch size set: Processes all items in a single batch:

```
    operations:
      mass_audience_rating_update: imdb
      # snip
```

Logs:

```
[INFO]     |========================================== Batch Updates ===========================================|
[INFO]     |                                                                                                    |
[INFO]     | Batch Audience Rating Update (1/4): 1 Movie updated to 5.9                                         |
[INFO]     | Batch Audience Rating Update (2/4): 1 Movie updated to 6.8                                         |
[INFO]     | Batch Audience Rating Update (3/4): 2 Movies updated to 7.1                                        |
[INFO]     | Batch Audience Rating Update (4/4): 1 Movie updated to 8.1                                         |
[INFO]     | Batch Critic Rating Update (1/2): 1 Movie updated to 0.8                                           |
[INFO]     | Batch Critic Rating Update (2/2): 1 Movie updated to 8.8                                           |
[INFO]     |                                                                                                    |
[INFO]     |====================================================================================================|
[INFO]     |                                        Done with Batch 1/1                                         |
[INFO]     |====================================================================================================|
```

### Batch size set to 100:

```
    operations:
      plex_bulk_edit_batch_size: 100
      mass_audience_rating_update: imdb
      # snip
```

Logs:

```
[INFO]     |========================================== Batch Updates ===========================================|
[INFO]     |                                                                                                    |
[INFO]     |                                                                                                    |
[INFO]     |====================================================================================================|
[INFO]     |                                        Done with Batch 1/7                                         |
[INFO]     |====================================================================================================|
[INFO]     |====================================================================================================|
[INFO]     |                                        Processing Batch 2/7                                        |
[INFO]     |====================================================================================================|

```

YozoraXCII:
I've asked @meisnate12 to review this for his approval, given this is a pretty significant change.

d3mystified:
Okay. @meisnate12 let me know what you think please.

meisnate12:
looks good i cleaned up the log statements and the generator. As long as when @YozoraXCII runs another test with the updated code and it passes this should be good to merge

miaLDN:
> For my very large library (10k+ movies) ratings updates would never finish (because something would cause kometa to break - reverse proxy, server instability, host instability, OOM etc). This change makes it way more reliable - I'm actually getting updates done now.

Hopefully the overhead of the logging and batch loop doesn't impact overall run times?

I have perhaps a similar number of movies to you; and the mass edits aren't too bad.  For me it's the TV series that takes a long long time.  It's only processing to season level.  It takes so long, that I rarely run it on TV shows now.

d3mystified:
It's not the logging that adds the delay but the plex operations. It used to be one large batch of plex operations, now it's N - so each one adds certain latency.

If it's too much for you, you can always adjust `plex_bulk_edit_batch_size` or disable it entirely (by not setting it).

PR #2822, opened 2025-09-25
avoid a logic issue with inflght reauth

chazlarson:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

During an inflight reauth, the request that triggered the reauth would then fail because we checked for a >=400 response code *after* successsfully reauthing the token.  This change wraps the 401 reauth with the check for >=400 so we reauth and/or report failure based on the one look at the response code.

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No


meisnate12:
didnt notice you made this and fixed it in another PR

PR #2911, opened 2026-01-02
New option `auto_sort_hub` to sort Recommendation Hubs by sort title

nwithan8:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

I rely on sort titles prefixed with special characters (e.g. `_`, `+` and `-`) to arrange the order of collections on the Collections tab in each library opaquely (independent of the visible collection name). For example, I put "trending" collections above "genre" collections above "studio" collections, etc.

It annoys me, however, that the Recommendation Hubs (the rows on the home screen and Recommendation tabs in each library) don't automatically respect the sort titles. Instead, if a new collection is generated, it's automatically placed at the bottom of the list, and I have to manually go in and drag the listing to its proper location.

For me, I generally want the Recommendation Hubs to mirror the order of the collections via sort titles. Therefore, I've introduced a new optional setting that will do just that: automatically rearrange all (visible) collections in the Recommendation Hubs listing so that it matches the sort title order.

## Related Issues [optional]

N/A

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/config/config.yml.template
+++ b/config/config.yml.template
@@ -76,6 +76,7 @@ settings:
   default_collection_order: release
   delete_below_minimum: true
   delete_not_scheduled: false
+  auto_sort_hubs: false
   run_again_delay: 2
   missing_only_released: false
   only_filter_missing: false
--- a/docs/config/settings.md
+++ b/docs/config/settings.md
@@ -136,6 +136,26 @@ The available setting attributes which can be set at each level are outlined bel
           asset_folders: true
         ```
 
+??? blank "`auto_sort_hubs` - Used to automatically sort Plex Recommendation Hubs by sort title order.<a class="headerlink" href="#auto-sort-hubs" title="Permanent link">¶</a>"
+
+    <div id="auto-sort-hubs" />When enabled, Kometa will automatically sort Plex Recommendation Hubs based on the sort titles of the collections within each library.
+
+    This will mirror the order of how collections are displayed within the Plex library view. Only collections that are marked as visible in Plex will be sorted.
+
+    **Attribute:** `auto_sort_hubs`
+
+    **Levels with this Attribute:** Global/Library
+
+    **Accepted Values:** `true` or `false`.
+
+    **Default Value:** `false`
+
+    ???+ example "Example"
+
+        ```yaml
+        settings:
+          auto_sort_hubs: true
+        ```
 
 ??? blank "`cache` - Used to control Kometa's cache database.<a class="headerlink" href="#cache" title="Permanent link">¶</a>"
 
--- a/docs/kometa/install/walkthroughs/kubernetes.md
+++ b/docs/kometa/install/walkthroughs/kubernetes.md
@@ -146,6 +146,7 @@ data:
       default_collection_order:
       delete_below_minimum: true
       delete_not_scheduled: false
+      auto_sort_hubs: false
       run_again_delay: 2
       missing_only_released: false
       only_filter_missing: false
--- a/json-schema/config-schema.json
+++ b/json-schema/config-schema.json
@@ -775,6 +775,10 @@
                     "description": "Used to delete collections not scheduled.\nIf a collection is skipped due to it not being scheduled, delete the collection.",
                     "type": "boolean"
                 },
+                "auto_sort_hubs": {
+                    "description": "Used to automatically sort Plex recommendation hubs after collections are updated, mirroring the sort title order.",
+                    "type": "boolean"
+                },
                 "run_again_delay": {
                     "description": "Used to control the number of minutes to delay running run_again collections.\nSet the number of minutes to delay running run_again collections after daily run is finished. For example, if a collection adds items to Sonarr/Radarr, the library can automatically re-run 'X' amount of time later so that any downloaded items are processed.",
                     "type": "integer",
--- a/json-schema/kitchen_sink_config.yml
+++ b/json-schema/kitchen_sink_config.yml
@@ -819,6 +819,7 @@ settings:                                           # Can be individually specif
   minimum_items: 1
   delete_below_minimum: true
   delete_not_scheduled: false
+  auto_sort_hubs: false
   run_again_delay: 1
   missing_only_released: true
   show_unconfigured: true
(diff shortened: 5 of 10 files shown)


YozoraXCII:
It might be more granular to be able to set `hub_priority` or something as an attribute in a collection definition rather than a singular setting? something like

```yaml
collections:
  My Collection:
    plex_all: true
    hub_priority: 1
  My Other Collection:
    plex_all: true
    hub_priority: 2
```

nwithan8:
> It might be more granular to be able to set `hub_priority` or something as an attribute in a collection definition rather than a singular setting? something like
> 
> ```yaml
> collections:
>   My Collection:
>     plex_all: true
>     hub_priority: 1
>   My Other Collection:
>     plex_all: true
>     hub_priority: 2
> ```

Unless I'm misunderstanding your recommendation, wouldn't you have to go through each collection and add a hub priority (or have some default "place it at the bottom" if not defined). And what would be the tiebreaker for two collections with the same priority?

I'm not knocking the idea, but that seems to introduce an entirely separate mechanism for sorting hubs. My intention was just to use the existing sortTitle method that Plex by default uses on collection pages (and therefore what I imagine a lot of people have already employed through "sort title" manipulation) and mirror that order onto the hubs, at least as a first pass of this concept.

YozoraXCII:
> > It might be more granular to be able to set `hub_priority` or something as an attribute in a collection definition rather than a singular setting? something like
> > ```yaml
> > collections:
> >   My Collection:
> >     plex_all: true
> >     hub_priority: 1
> >   My Other Collection:
> >     plex_all: true
> >     hub_priority: 2
> > ```
> 
> Unless I'm misunderstanding your recommendation, wouldn't you have to go through each collection and add a hub priority (or have some default "place it at the bottom" if not defined). And what would be the tiebreaker for two collections with the same priority?
> 
> I'm not knocking the idea, but that seems to introduce an entirely separate mechanism for sorting hubs. My intention was just to use the existing sortTitle method that Plex by default uses on collection pages (and therefore what I imagine a lot of people have already employed through "sort title" manipulation) and mirror that order onto the hubs, at least as a first pass of this concept.

Yes that's what I'm proposing I guess, so that users have granular control over the hubs rather than just sorting by sort title. It's something that's been asked for on more than one occasion. 

If two collections have the same priority, it could then fall to sort title.

If `auto_sort_hub` is to go ahead, it should probably be a per-library setting since users may not want it on all libraries

YozoraXCII:
> > > It might be more granular to be able to set `hub_priority` or something as an attribute in a collection definition rather than a singular setting? something like
> > > ```yaml
> > > collections:
> > >   My Collection:
> > >     plex_all: true
> > >     hub_priority: 1
> > >   My Other Collection:
> > >     plex_all: true
> > >     hub_priority: 2
> > > ```
> > 
> > Unless I'm misunderstanding your recommendation, wouldn't you have to go through each collection and add a hub priority (or have some default "place it at the bottom" if not defined). And what would be the tiebreaker for two collections with the same priority?
> > 
> > I'm not knocking the idea, but that seems to introduce an entirely separate mechanism for sorting hubs. My intention was just to use the existing sortTitle method that Plex by default uses on collection pages (and therefore what I imagine a lot of people have already employed through "sort title" manipulation) and mirror that order onto the hubs, at least as a first pass of this concept.
> 
> Yes that's what I'm proposing I guess, so that users have granular control over the hubs rather than just sorting by sort title. It's something that's been asked for on more than one occasion. 
> 
> If two collections have the same priority, it could then fall to sort title.
> 
> If `auto_sort_hub` is to go ahead, it should probably be a per-library setting since users may not want it on all libraries

@nwithan8 thoughts on the above?

nwithan8:
@YozoraXCII sounds reasonable. Admittedly, I'm not immediately sure how to do per-collection settings, and specifically how to extract that setting and collate them all together, but I'll explore the code base and try to figure it out.

badja-dev:
Picking this one up

badja-dev:
@nwithan8 

Since I can't push to your branch directly, I'm happy to open a PR to `nwithan8:hub_move_sync` with the fixes if that works for you?
After reviewing the code, a few things I'd want to address before this merges:

Call site timing - `sort_collection_hubs_by_sort_title()` is called inside run_collection(), which runs once per collection file. A library with multiple collection files triggers the sort multiple times. Should move up to `run_libraries()` after all collection files for the library are processed.
N+1 API calls - `is_collection_visible()` makes one API call per collection to check hub visibility. The existing `manage_recommendations()` already shows the right pattern - a single call to `/hubs/sections/{key}/manage` returns everything at once. Worth refactoring to use that.
Error handling & logging - no try/except and no log output inside the sort function, which is inconsistent with how Kometa handles everything else at this level.
`from typing import List` - Kometa is Python 3.10+, so `list[Collection]` works natively.
CHANGELOG - needs updating before merge.

None of this is a blocker on the concept - the approach is sound. Let me know if you want me to open a PR with these fixes, or if you'd prefer to address them yourself.

YozoraXCII:
> @nwithan8
> 
> Since I can't push to your branch directly, I'm happy to open a PR to `nwithan8:hub_move_sync` with the fixes if that works for you? After reviewing the code, a few things I'd want to address before this merges:
> 
> Call site timing - `sort_collection_hubs_by_sort_title()` is called inside run_collection(), which runs once per collection file. A library with multiple collection files triggers the sort multiple times. Should move up to `run_libraries()` after all collection files for the library are processed. N+1 API calls - `is_collection_visible()` makes one API call per collection to check hub visibility. The existing `manage_recommendations()` already shows the right pattern - a single call to `/hubs/sections/{key}/manage` returns everything at once. Worth refactoring to use that. Error handling & logging - no try/except and no log output inside the sort function, which is inconsistent with how Kometa handles everything else at this level. `from typing import List` - Kometa is Python 3.10+, so `list[Collection]` works natively. CHANGELOG - needs updating before merge.
> 
> None of this is a blocker on the concept - the approach is sound. Let me know if you want me to open a PR with these fixes, or if you'd prefer to address them yourself.

I think we were talking about moving away from this approach and instead using a `hub_priority` attribute at the Collection definition level, like this:
```
collections:
  My Collection:
    plex_all: true
    hub_priority: 1
  My Other Collection:
    plex_all: true
    hub_priority: 2
```

The idea being that once all collections are processed, it would then use the priority to rearrange the Hub items in that particular order

YozoraXCII:
Closing as Superceded

Issue #3054, opened 2026-04-05
Letterboxd popular this week list returns no items


hanichols:
### Version Number

2.3.1.4

### What branch are you on?

nightly

### Describe the Bug

The letterboxd popular list at https://letterboxd.com/films/popular/this/week/ returns no items during validation which previously worked. Other letterboxd lists under the https://letterboxd.com/<username>/list/<listname> heading work as expected.

### Relevant Collection/Overlay/Playlist Definition

```yml
Popular:
    sync_mode: sync
    collection_order: custom
    radarr_add_missing: true
    letterboxd_list_details:
      url:  https://letterboxd.com/films/popular/this/week/
      limit: 75
    summary: Popular Films This Week on Letterboxd
```

### Logs

https://gist.github.com/hanichols/a3baffeb59a568c9bef1ae9d9275bbda

chazlarson:
Letterboxd returning `403 forbidden` so probably some anti-scraping stuff:
```
|============================= Validating Popular This Week Attributes ==============================|
| http://192.168.1.11:32400 "GET /library/sections/1/all?includeGuids=1&title=Popular%20This%20Week&type=18 HTTP/1.1" 200 324 |
|                                                                                                    |
| Looking for: Popular This Week                                                                     |
|                                                                                                    |
| Validating Method: letterboxd_list                                                                 |
| Value: https://letterboxd.com/films/popular/this/week/                                             |
| URL: https://letterboxd.com/films/ajax/popular/this/week/                                          |
| Starting new HTTPS connection (1): letterboxd.com:443                                              |
| https://letterboxd.com:443 "GET /films/ajax/popular/this/week/ HTTP/1.1" 403 None                  |
| Starting new HTTPS connection (1): letterboxd.com:443                                              |
| https://letterboxd.com:443 "GET /films/ajax/popular/this/week/ HTTP/1.1" 403 None                  |
| Collection Warning: https://letterboxd.com/films/popular/this/week/ returned no items during validation |
|                                                                                                    |
| Validation Successful                                                                              |
|                                                                                                    |
|=============================== Running Popular This Week Collection ===============================|
|                                                                                                    |
| Sync Mode: sync                                                                                    |
|                                                                                                    |
| Builder: letterboxd_list: {'url': 'https://letterboxd.com/films/popular/this/week/', 'limit': 0, 'note': None, 'rating': None, 'year': None} |
|                                                                                                    |
[...]

PR #3083, opened 2026-04-24
Fix Letterboxd list/watchlist parser for LazyPoster markup

flakypickle21:
## What type of PR is this?

## Description

Letterboxd migrated list and watchlist pages to a React-based `LazyPoster` component in April 2026. The `<div data-film-id="...">` nodes `modules/letterboxd.py` matches no longer exist; film data is now on:

```html
<div class="react-component"
     data-component-class="LazyPoster"
     data-item-slug="interstellar"
     data-item-link="/film/interstellar/"
     data-item-name="Interstellar (2014)"
     data-postered-identifier='{"lid":"4VZ8","uid":"film:117621","type":"film","typeName":"film"}'
     ...>
```

As a result every `letterboxd_list` / `letterboxd_user_films` / `letterboxd_user_reviews` / watchlist builder fails with `Letterboxd Error: No List Items found in {...}` during `get_tmdb_ids`, and emits `returned no items during validation` in `validate_letterboxd_lists`.

This PR updates `_parse_page` and `_parse_user_page` to match the new element via `//div[@data-item-slug]` and extract:

- **slug** from `data-item-link` (falls back to the legacy `data-target-link`, which still appears on the same element)
- **numeric film id** from the `data-postered-identifier` JSON (`uid: "film:<id>"`). The id format is identical to the old `data-film-id` value, so existing `letterboxd_map` cache rows keep matching — no forced re-scrape of TMDb IDs on upgrade
- **year** via regex from `data-item-name` (e.g. `"Interstellar (2014)"`)

Rating / note are left as `None` on list and watchlist items — the new poster markup has no `<article>` ancestor (watchlist items sit inside `<li class="griditem">`, list items inside `<li class="posteritem numbered-list-item">`), so the legacy sibling xpaths for those fields match nothing. Restoring them depends on the diary / `/<user>/films/` pages, which are currently returning `403 Forbidden` even through `cloudscraper` (the underlying problem behind #3054). That is a transport-level issue and warrants a separate pass.

`_tmdb()` is unchanged — `/film/<slug>/` pages still expose `data-track-action="TMDb"` when fetched via `cloudscraper`, and the existing xpath resolves correctly.

### Testing

Exercised `_parse_list` against six real URLs inside a running Kometa 2.3.1.4 (Docker master) container after swapping in the patched `modules/letterboxd.py`. Before the fix, all six return 0 items with `No List Items found`. After:

| URL shape | Items before | Items after | First item |
|---|---:|---:|---|
| `/<user>/watchlist/` | 0 | 159 | `id=41361 year=1979 /film/all-that-jazz/` |
| `/ellefnning/list/for-when-you-want-to-feel-something/` | 0 | 186 | `id=730798 year=2022 /film/cha-cha-real-smooth/` |
| `/cinemaparadisa/list/comfort-movies/` | 0 | 183 | `id=95113 year=2014 /film/the-grand-budapest-hotel/` |
| `/official/list/top-250-films-with-the-most-fans/` | 0 | 250 | `id=117621 year=2014 /film/interstellar/` |
| `/dave/list/imdb-top-250/` | 0 | 250 | `id=51778 year=1994 /film/the-shawshank-redemption/` |
| `/alexanderh/list/letterboxd-one-million-watched-club/` | 0 | 748 | `id=636761 year=2023 /film/evil-dead-rise/` |

Cache-compat spot check: `id=51778` for Shawshank matches the value historically stored under `data-film-id` in the old markup, so existing `letterboxd_map` rows keyed on that id remain valid.

## Related Issues [optional]

- Related Issue #3054 — same family of breakage. This PR fixes the parser-level manifestation affecting `/<user>/list/` and `/<user>/watchlist/`. The original `/films/popular/this/week/` scenario in #3054 is still `403 Forbidden` through `cloudscraper`, so this PR is **not** a full close of that issue.

- [x] No (internal parser change; user-facing docs remain accurate)

- [x] Yes

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,7 +28,7 @@ Updates to Trakt and MAL flow pages
 # Defaults
 
 # Bug Fixes
-Update Letterboxd list/watchlist parser for the React-based `LazyPoster` markup that replaced `data-film-id`/`data-target-link` nodes, restoring collections built from `letterboxd_list` and `letterboxd_user_films`/`letterboxd_user_reviews` on watchlists and user lists.
+Update Letterboxd list/watchlist parser for the React-based `LazyPoster` markup, which replaced the `data-film-id` nodes the `letterboxd_list`, `letterboxd_user_films`, and `letterboxd_user_reviews` builders relied on.
 Catch KeyError on Plex NFO build.
 Fixing "list index out of range" error when processing imdb charts #3006
 TMDB language now reflected in mass_poster_update; cache extension
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -28,6 +28,7 @@ Updates to Trakt and MAL flow pages
 # Defaults
 
 # Bug Fixes
+Update Letterboxd list/watchlist parser for the React-based `LazyPoster` markup that replaced `data-film-id`/`data-target-link` nodes, restoring collections built from `letterboxd_list` and `letterboxd_user_films`/`letterboxd_user_reviews` on watchlists and user lists.
 Catch KeyError on Plex NFO build.
 Fixing "list index out of range" error when processing imdb charts #3006
 TMDB language now reflected in mass_poster_update; cache extension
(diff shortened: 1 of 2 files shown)


chazlarson:
LGTM

PR #3100, opened 2026-05-10
Fix relative paths to resolve against config file directory (#3092)

### LINKED ISSUE #3092: Relative paths in config resolve against CWD instead of config file directory when using custom config path
## Description

When the config file path is changed from the default (via the `-c` flag or the equivalent environment variable), relative paths defined inside the config (e.g. `collection_files`, `metadata_files`, `overlay_files`, `folder`, `asset_directory`) are still resolved against the **current working directory (CWD)** rather than the directory of the specified config file.

This means the feature works correctly only when Kometa is run from the same directory as the default `config/` folder. As soon as a custom config path is used, all relative paths inside that config silently fail to resolve.

## Steps to Reproduce

1. Place your `config.yml` in a non-default location, e.g. `/custom/path/config.yml`
2. Use a relative path for a collection/metadata/overlay file inside that config, e.g.:
   ```yaml
   collection_files:
     - file: my_collections.yml
   ```
3. Run Kometa with `-c /custom/path/config.yml`
4. Kometa reports the file is not found, even though `my_collections.yml` exists next to `config.yml`

## Root Cause

There are two locations where relative paths are incorrectly resolved against CWD:

### 1. `modules/util.py` — `load_files()` (line ~340)

The function has no `config_dir` parameter. Three code paths inside it resolve relative paths against CWD:

- **`folder` key** (line ~376): `os.path.isdir(file["folder"])` — relative folder paths check against CWD
- **`asset_directory` key** (line ~399): `os.path.exists(asset_path)` — relative asset paths check against CWD
- **Plain string paths** (line ~428):
  ```python
  if os.path.exists(file):   # resolves relative to CWD, not config dir
      files.append(("File", file, {}, None))
  ```

### 2. `modules/meta.py` — `load_file()` (line ~152)

```python
content_path = os.path.abspath(file_path)
```

`os.path.abspath()` with no explicit base always resolves relative to CWD, not the config file's directory.

## Expected Behaviour

All relative paths in the config should resolve relative to the directory containing the config file, regardless of the CWD or how the config path was specified.

## Current Workaround

Use absolute paths in the config instead of relative ones.

## Suggested Fix

1. Add a `config_dir` parameter to `util.load_files()` and use `os.path.join(config_dir, path)` when encountering relative paths for `file`, `folder`, and `asset_directory` entries.
2. Update the three `util.load_files()` call sites in `modules/config.py` (lines ~1609, ~1629, ~1666) to pass `self.default_dir`.
3. In `meta.py:load_file()`, resolve relative `File`-type paths against the config directory rather than CWD.

StarshipSuperjam:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

When Kometa runs with `-c <custom>/config.yml`, relative paths in `collection_files`, `metadata_files`, `overlay_files`, `playlist_files`, `image_files`, `external_templates`, `style_file`, `folder`, and `asset_directory` currently resolve against the current working directory instead of the directory containing the config file. This silently breaks the default Docker setup (WORKDIR=`/`, config at `/config/config.yml`), cron/launchd-driven jobs, and any other custom-`-c` invocation. After this change those paths resolve relative to the config file's directory; absolute paths are unaffected.

**Implementation**
- `modules/util.py`: new `resolve_path(config_dir, path)` helper and a new optional `config_dir` kwarg on `load_files()`. The `folder` (isdir + glob), `asset_directory` (exists), and plain-string file (exists) branches route through the helper.
- `modules/meta.py`: `DataFile.load_file()` resolves `file_path` via `util.resolve_path(self.config.default_dir, file_path)` before computing `content_path` / `dir_path`, which cascades to the translation `os.listdir(dir_path)` and the per-language file path.
- All seven `util.load_files(...)` call sites pass `config_dir=` from the `default_dir` already in scope: `modules/config.py` (playlist_files, collection_files, metadata_files, overlay_files, image_files) and `modules/meta.py` (external_templates, style_file).

**Backward compatibility**
`config_dir=None` default preserves the prior CWD-relative behavior for any external caller of `load_files()`. Absolute paths and Kometa Defaults pass through (`os.path.isabs` short-circuits the helper).

**Testing**
- Direct-function probe driving `util.load_files()` from `cd /` over six fixture cases (scalar relative, dict `folder` relative, dict `folder` absolute regression, dict `file`, `asset_directory`, no-`config_dir` backward-compat). Pre-fix the scalar-relative case reproduced `Config Error: Path not found` and returned `[]`; post-fix it returns the resolved absolute tuple. See follow-up comment for the script and output, plus Windows path-semantics validation via `ntpath`.
- `pyspelling -c .spellcheck.yml` clean.
- `validate-pull` and `docs/readthedocs.org:kometa` checks PASS.

## Related Issues [optional]

- Closes #3092

- [X] Yes — a "Relative path resolution" note in `docs/config/files.md` above the `file` / `folder` blocks.
- [ ] No

- [X] Yes — entry under `# Bug Fixes`.
- [ ] No


chazlarson:
Is this tested on Windows?

chazlarson:
Also, the PR does not follow the standard requested template.

StarshipSuperjam:
@chazlarson — apologies for missing the PR template on the initial submission; that was an oversight on my part. I've just updated the PR description above to follow `.github/pull_request_template.md`.

On the Windows question: I don't have a Windows machine to drive a full end-to-end run, so let me cover what I can.

**Analysis.** The fix only adds calls to `os.path.isabs`, `os.path.join`, and `os.path.abspath`, all of which dispatch to OS-specific implementations from `posixpath` / `ntpath`. No hardcoded separators were introduced. Concretely, on Windows:
- `os.path.isabs` returns True for `C:\foo`, `C:/foo`, `\\server\share\foo`, and `\foo` — so absolute, mixed-slash, UNC, and rooted paths all pass through `resolve_path` unchanged.
- `os.path.join(config_dir, rel)` produces a valid Windows path whether `rel` uses `\`, `/`, or mixes them; Windows accepts both separators throughout, and `os.path.exists` / `os.path.isdir` / `os.listdir` handle either.
- `os.path.abspath` normalizes `..` segments and separators using the host OS's rules.

**Cross-platform validation.** Python ships `ntpath`, which implements Windows path semantics regardless of host OS. The probe below mirrors `util.resolve_path` byte-for-byte except it imports `ntpath` in place of `os.path`, so the algorithm can be exercised exactly as it would run on Windows from any developer's machine.

```python
# /tmp/3092-windows-probe.py
"""Validate resolve_path against Windows path semantics, host-independent."""
import ntpath

def resolve_path_nt(config_dir, path):
    if not path or not config_dir or ntpath.isabs(path):
        return path
    return ntpath.abspath(ntpath.join(config_dir, path))

CFG = r"C:\opt\kometa"
cases = [
    ("relative, backslashes",       r"sub\coll.yml"),
    ("relative, forward slashes",   r"sub/coll.yml"),
    ("relative with .. segment",    r"..\other\coll.yml"),
    ("absolute, drive + backslash", r"C:\data\coll.yml"),
    ("absolute, drive + fwd slash", r"C:/data/coll.yml"),
    ("UNC path",                    r"\\server\share\coll.yml"),
    ("rooted, no drive",            r"\foo\bar.yml"),
    ("empty path",                  ""),
]
for name, inp in cases:
    print(f"{name:30}  {inp!r:35}  ->  {resolve_path_nt(CFG, inp)!r}")
```

Output (run on macOS, exercises Windows semantics):

```
relative, backslashes           'sub\\coll.yml'                      ->  'C:\\opt\\kometa\\sub\\coll.yml'
relative, forward slashes       'sub/coll.yml'                       ->  'C:\\opt\\kometa\\sub\\coll.yml'
relative with .. segment        '..\\other\\coll.yml'                ->  'C:\\opt\\other\\coll.yml'
absolute, drive + backslash     'C:\\data\\coll.yml'                 ->  'C:\\data\\coll.yml'
absolute, drive + fwd slash     'C:/data/coll.yml'                   ->  'C:/data/coll.yml'
UNC path                        '\\\\server\\share\\coll.yml'        ->  '\\\\server\\share\\coll.yml'
rooted, no drive                '\\foo\\bar.yml'                     ->  '\\foo\\bar.yml'
empty path                      ''                                   ->  ''
```

Every case behaves as expected on Windows: relatives resolve under `config_dir` with backslash normalization, mixed forward/back slashes are preserved (Windows accepts both), `..` segments collapse correctly, and absolute / UNC / rooted paths pass through unchanged.

**End-to-end on a Windows box.** If you (or anyone reading this) wants a full reproducer rather than the algorithmic check above, the fixture is straightforward:

```cmd
:: Run from a CWD that isn't C:\opt\kometa
mkdir C:\opt\kometa\sub
echo collections:>  C:\opt\kometa\sub\coll.yml
echo:   Smoke Test:>> C:\opt\kometa\sub\coll.yml
echo:     smart_filter: { sort_by: random, limit: 1 }>> C:\opt\kometa\sub\coll.yml

:: Minimal config.yml at C:\opt\kometa\config.yml with:
::   libraries:
::     Movies:
::       collection_files:
::         - file: sub\coll.yml
::         - folder: sub
:: plus real plex / tmdb creds

cd C:\
python C:\Kometa\kometa.py -c C:\opt\kometa\config.yml -r --no-countdown
```

Before this PR: `Config Error: Path not found: sub\coll.yml` and `Config Error: Folder not found: sub`.
After: file/folder load normally.

Happy to incorporate any further changes you'd like.

YozoraXCII:
Okay so, as someone who doesn't really understand this - I ran a quick test.

Before implementing this PR:

Kometa install directory: `D:\Repositories\Kometa`
config.yml has:
```
libraries:
  Movies:
    collection_files:
    - file: config/test.yml
```
test.yml's full filepath: `D:\Repositories\Kometa\config\test.yml`

I ran `python .\kometa.py -c /config.yml -r` which failed with `Config Error: Configuration file (config.yml) not found at D:\config.yml`

I placed the config.yml in `D:\` and ran Kometa, which worked:
```
|====================================================================================================|
|                                        Movies Configuration                                        |
|====================================================================================================|
|                                                                                                    |
| Connecting to Movies Library...                                                                    |
|                                                                                                    |
| Reading in Collection Files                                                                        |
|                                                                                                    |
| Reading file: config/test.yml                                                                      |
```

I then implemented your 3 file changes (config.py, meta.py, util.py) and re-ran `python .\kometa.py -c /config.yml -r`, which failed with:
`File Error: File does not exist D:\config\test.yml`

I don't know if we are saying that this is an "expected failure" because the root of `config` is now `D:\config` rather than `D:\Repositories\Kometa\config`

chazlarson:
That seems expected.

What this PR seems to be addressing would be the opposite case.

Your config is at `D:\Repositories\Kometa\config\config.yml`

It contains a reference like:
```yaml
libraries:
  TV Shows:
    overlay_files:
    - file: config/cool-overlays.yml
```
If you do this:
```
d:
cd Repositories\Kometa\
python kometa.py
```
All is well, since `config/cool-overlays.yml` is found at `D:\Repositories\Kometa\config\cool-overlays.yml`

However, if you do this:
```
d:
cd \
python Repositories\Kometa\kometa.py --config Repositories\Kometa\config\config.yml
```
then `config/cool-overlays.yml` can't be found because Kometa will look for it at `d:\config\cool-overlays.yml`

This PR claims to address that and interpret this:
```
    - file: config/cool-overlays.yml
```
relative to the directory that contains the `config.yml`, not relative to the current working directory where you invoked Kometa.

YozoraXCII:
> That seems expected.
> 
> What this PR seems to be addressing would be the opposite case.
> 
> Your config is at `D:\Repositories\Kometa\config\config.yml`
> 
> It contains a reference like:
> 
> ```yaml
> libraries:
>   TV Shows:
>     overlay_files:
>     - file: config/cool-overlays.yml
> ```
> 
> If you do this:
> 
> ```
> d:
> cd Repositories\Kometa\
> python kometa.py
> ```
> 
> All is well, since `config/cool-overlays.yml` is found at `D:\Repositories\Kometa\config\cool-overlays.yml`
> 
> However, if you do this:
> 
> ```
> d:
> cd \
> python Repositories\Kometa\kometa.py --config Repositories\Kometa\config\config.yml
> ```
> 
> then `config/cool-overlays.yml` can't be found because Kometa will look for it at `d:\config\cool-overlays.yml`
> 
> This PR claims to address that and interpret this:
> 
> ```
>     - file: config/cool-overlays.yml
> ```
> 
> relative to the directory that contains the `config.yml`, not relative to the current working directory where you invoked Kometa.

So this still doesn't work quite as expected I believe, but maybe that's a "me" problem.

`D:\Repositories\Kometa\config\config.yml` and `D:\Repositories\Kometa\config\test.yml`
```
libraries:
  Movies:
    collection_files:
    - file: config/test.yml
```

```
d:
cd \
python .\Repositories\Kometa\kometa.py --config .\Repositories\Kometa\config\config.yml -r
```
results in
```
File Error: File does not exist D:\Repositories\Kometa\config\config\test.yml
```

so it's looking for a second `config` folder

chazlarson:
That looks like a bug in this implementation.

Perhaps only tested on files that aren't in the config dir?

In my experience, this:
```yaml
libraries:
  Movies:
    collection_files:
      - file: config\my_file.yml
```
Is far more common than
```yaml
libraries:
  Movies:
    collection_files:
      - file: something_else\my_file.yml
```
If this change works with the latter but not with the former, that's going to break most people's setups.

YozoraXCII:
Seems to still be broken
`D:\Repositories\Kometa\files\test.yml`
```
libraries:
  Movies:
    collection_files:
    - file: files/test.yml
```
```
(venv) PS D:\> python .\Repositories\Kometa\kometa.py --config .\Repositories\Kometa\config\config.yml -r
```
`File Error: File does not exist D:\Repositories\Kometa\config\files\test.yml`

Which I suppose makes some sense, as the path would be relative to the folder where config.yml is located, right? I suspect this may break some people's setup though as it seems like a change in behaviour maybe?


if I hardcode the location then yes it works, but I suspect most people aren't doing this:
```
libraries:
  Movies:
    collection_files:
    - file: D:\Repositories\Kometa\files\test.yml
```

YozoraXCII:
@StarshipSuperjam any thoughts on the above?

StarshipSuperjam:
@YozoraXCII @chazlarson — thanks for the careful testing, and you're both right; the implementation is wrong as written.

**Root cause:** `default_dir` is the folder that *contains* `config.yml`, which in a normal install is `…/Kometa/config/` itself. So resolving a relative `file: config/test.yml` against it produces `…/config/config/test.yml` — the doubling you saw. And as the `files/test.yml` case shows, the deeper problem is that people write relative paths against the **Kometa root / CWD**, not against the config dir, so switching the base unconditionally changes behavior for existing setups. That's exactly the regression you flagged.

I think the fix that satisfies #3092 without breaking anyone is **CWD-first with a config-dir fallback**: resolve each relative path against the CWD exactly as today, and only if that doesn't resolve, try the config dir.

- Every setup that works today is unchanged — `config/foo.yml` / `files/foo.yml` run from the Kometa root still resolve against the root, no regression.
- #3092 still gets fixed: running from a foreign CWD with files next to a custom `config.yml`, CWD resolution fails and the config-dir fallback picks them up.
- The only cases left unfixed are inherently ambiguous (foreign CWD + root-relative `config/foo.yml`), where no base can be inferred without guessing the install root.

Before I push that — does that match the semantics you'd want? Or would you rather I close this and leave resolution CWD-only? Happy to go either way; your call on what fits the project best.

YozoraXCII:
> @YozoraXCII @chazlarson — thanks for the careful testing, and you're both right; the implementation is wrong as written.
> 
> **Root cause:** `default_dir` is the folder that _contains_ `config.yml`, which in a normal install is `…/Kometa/config/` itself. So resolving a relative `file: config/test.yml` against it produces `…/config/config/test.yml` — the doubling you saw. And as the `files/test.yml` case shows, the deeper problem is that people write relative paths against the **Kometa root / CWD**, not against the config dir, so switching the base unconditionally changes behavior for existing setups. That's exactly the regression you flagged.
> 
> I think the fix that satisfies #3092 without breaking anyone is **CWD-first with a config-dir fallback**: resolve each relative path against the CWD exactly as today, and only if that doesn't resolve, try the config dir.
> 
> * Every setup that works today is unchanged — `config/foo.yml` / `files/foo.yml` run from the Kometa root still resolve against the root, no regression.
> * [Relative paths in config resolve against CWD instead of config file directory when using custom config path #3092](https://github.com/Kometa-Team/Kometa/issues/3092) still gets fixed: running from a foreign CWD with files next to a custom `config.yml`, CWD resolution fails and the config-dir fallback picks them up.
> * The only cases left unfixed are inherently ambiguous (foreign CWD + root-relative `config/foo.yml`), where no base can be inferred without guessing the install root.
> 
> Before I push that — does that match the semantics you'd want? Or would you rather I close this and leave resolution CWD-only? Happy to go either way; your call on what fits the project best.

@chazlarson thoughts?

chazlarson:
I don't understand this assertion:

> This silently breaks the default Docker setup (WORKDIR=/, config at /config/config.yml), cron/launchd-driven jobs, and any other custom--c invocation

I run all three of these things regularly and do not have any of the issues described here.

It feels like a lot of work to avoid putting a `cd /path/to/kometa` in a script as opposed to just documenting clearly that relative paths are relative to the CWD.

StarshipSuperjam:
Sorry. I'm AFK for a few weeks and won't be able to come back to this any time soon. Closing the PR.

PR #3119, opened 2026-05-18
perf: pre-compute frozensets for method-validation hot path in builder.py

Gilmoursa:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

In the `CollectionBuilder` method-validation loop, two checks were creating a **new list on every call** by concatenating module-level constants at runtime:

```python
# Before — allocates and linearly scans a ~500-item list on every attribute validated
method_name in all_builders + plex.searches
method_name in all_builders + smart_url_invalid
```

Both operands are fixed at import time, so the concatenation result never changes. This PR pre-computes them once as module-level `frozenset`s:

```python
all_builders_and_searches = frozenset(all_builders + plex.searches)
all_builders_and_smart_url_invalid = frozenset(all_builders + smart_url_invalid)
```

The two hot-path lines are updated to use these sets instead.

**Impact:** The validation loop runs for every attribute of every collection on every Kometa run. A typical run with 100 collections × ~20 attributes hits these checks ~2,000 times. Each call previously allocated and scanned a list of ~500+ items (O(n)); it is now a single O(1) hash lookup with no allocation.

Zero behavioral change — this is a pure performance optimization.

## Related Issues

- Related Issue #

- [X] No — internal implementation change, no user-facing behavior affected

- [X] Yes

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ Bump setuptools to 82.0.1
 
 # Performance
 Use set lookup instead of list scan when checking collection membership in `add_to_collection` and `run_collections_again`, reducing complexity from O(n×m) to O(n).
+Pre-compute `frozenset`s for `all_builders + plex.searches` and `all_builders + smart_url_invalid` used in the method-validation hot path, eliminating per-call list allocation and reducing membership-test complexity from O(n) to O(1).
 
 # New Features
 Added `plex_csm` and `plex_csm0` sources to `mass_content_rating_update` (reads CSM age rating from Plex's own cached `commonSenseMedia.ageRatings`, no external API key required).
--- a/modules/builder.py
+++ b/modules/builder.py
@@ -476,6 +476,8 @@ smart_url_invalid = (
     + radarr_details
     + sonarr_details
 )
+all_builders_and_searches = frozenset(all_builders + plex.searches)
+all_builders_and_smart_url_invalid = frozenset(all_builders + smart_url_invalid)
 custom_sort_builders = [
     "plex_search",
     "plex_watchlist",
@@ -1498,7 +1500,7 @@ class CollectionBuilder:
             logger.debug(f"Validating Method: {method_key}")
             logger.debug(f"Value: {method_data}")
             try:
-                if method_data is None and method_name in all_builders + plex.searches and method_final not in none_builders:
+                if method_data is None and method_name in all_builders_and_searches and method_final not in none_builders:
                     raise Failed(f"{self.Type} Error: {method_final} attribute is blank")
                 elif method_data is None and method_final not in none_details:
                     logger.warning(f"Collection Warning: {method_final} attribute is blank")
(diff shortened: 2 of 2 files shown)


chazlarson:
How are all these being tested?  Are you actually confirming that the user-visible behavior is the same before and after?

Gilmoursa:
**Testing approach:**

Since both replaced expressions (`all_builders + plex.searches` and `all_builders + smart_url_invalid`) are composed entirely of module-level constants that are never mutated after import, the correctness guarantee is provable by set-equality rather than runtime observation.

I wrote and ran 7 unit tests directly against the changed module that verify:
1. Both new names are `frozenset` instances
2. Each frozenset contains **exactly** the same members as the original list expression (`set(original) == frozenset`)
3. `item in frozenset` returns the same result as `item in original_list` for every item in `all_builders`, `plex.searches`, and `smart_url_invalid`

All 7 passed. The `in` operator on a `frozenset` uses hash-based equality, semantically identical to a list scan, just O(1) instead of O(n).

I also attempted a live Kometa dry-run against my local Plex instance but hit a session-layer connection issue when running outside Docker. Happy to retest on the Unraid host directly if you'd prefer an end-to-end confirmation.

chazlarson:
The thing is that Kometa users don't run unit tests.  They run the tool.  In all likelihood this sort of internal thing is fine, but personally I like to see that the actual tool continues to work as expected as opposed to just the unit tests.

It would be preferable to catch any unexpected follow-on effects *before* everyone in the Discord starts complaining.  But then I'm just a cranky old man who used to ship software on CDs. ;)

Gilmoursa:
Haha, I definitely understand that. Let me get you some hard data and get back to you on this.

Gilmoursa:
Ran a targeted dry run against my live setup (Kometa in Docker on Unraid) using `--run-collections "HBO Max Movies"`:

- All collection files loaded successfully
- HBO Max Movies collection processed without errors
- Run completed cleanly in 0:00:47

No unexpected behaviour from the frozenset change. Happy to share the full log if useful.

Gilmoursa:
Following up as the previous run was against the unmodified Kometa image. I re-ran the dry run with the PR branch (`perf/frozenset-method-validation`) mounted directly, using the modified `modules/builder.py`:

- All collection files loaded successfully
- HBO Max Movies collection processed without errors
- Run completed cleanly in 0:00:47

Output and runtime are identical to the baseline. No unexpected behaviour from the frozenset change.

Gilmoursa:
After running comparisons against both a small collection (4 items) and a larger one (889 items), the runtime was identical between the original image and the PR branch in both cases. The bottleneck in practice is Plex API calls and metadata processing, not list membership checks, so the frozenset optimization doesn't produce a measurable improvement for real workloads.

Given that, I don't think this change is worth merging. Closing the PR. Thanks for pushing back on the testing question, it led to a more honest answer.

PR #3129, opened 2026-05-24
fix: downgrade stale TVDb 4xx responses to debug-level log

SturmB:
## What type of PR is this?

- [X] Bug Fix (non-breaking change which fixes an issue)

## Description

Closes #3047. The reporter saw ~137 `TVDb Error: No Series found for TVDb ID: ...` warnings per run from default collection files (franchise, streaming, network, etc.) and asked whether the IDs could be audited.

The fix path turned out to be different from auditing default files. I scanned the entire Kometa tree (every `.yml/.yaml/.json/.py/.md/.txt`) for any of the 136 unique stale IDs — **zero matches**. Same result for the remote data sources Kometa fetches at runtime: `Anime-IDs/anime_ids.json` matches just 1/136, `Mediastingers/stingers.yml` matches 0/136 (keyed by TMDb ID, not TVDb).

The IDs originate from TMDb's `external_ids` data for TMDb shows that the defaults reference. Kometa's `Convert.tmdb_to_tvdb()` reads those values, then later uses them when iterating missing shows (`builder.py` ~4396 / ~5076) and evaluating `tvdb_*` filters (`builder.py` ~4287). When the cached TMDb→TVDb mapping points at a series that no longer exists on thetvdb.com, `TVDbObj.__init__` raises `Failed` and the caller logs `logger.error(e)` — which the reporter says clutters logs and triggers Notifiarr/Discord failure notifications.

These IDs are not user-controlled, so an error is misleading. This PR:

- Adds `modules.tvdb.NotFound` (subclass of `Failed`) for HTTP 4xx responses from thetvdb.com dereferrer endpoints. 5xx still raises plain `Failed` (transient TVDb outage / parse error).
- Raises `NotFound` from `TVDb.get_request` on 4xx and propagates it through `TVDbObj.__init__`.
- In the three `builder.py` sites that scan IDs the user did not type (`missing_shows`, `run_again_shows`, `tvdb_f` filter), catches `NotFound` separately and logs at debug.

User-typed builders like `tvdb_show_details: <id>` are unchanged — those propagate `NotFound`/`Failed` to the outer handler so the user still gets the loud error if they typed an invalid ID.

Tested:

- `pytest tests/test_tvdb.py -v` → 4/4 passing (NotFound subclass, 4xx mapping, 5xx mapping, propagation).
- Full `pytest tests/` → no regressions vs unmodified `nightly` (the same 6 pre-existing failures in `test_builder.py` and `test_letterboxd.py` reproduce on a clean checkout).

[...]

Issue #3047, opened 2026-03-31
~137 stale TVDb series IDs in defaults produce 'No Series found' errors


SturmB:
## Description

Running Kometa 2.3.0 (also reproduced on nightly), the default collection files produce ~137 `TVDb Error: No Series found for TVDb ID: XXXXXX` warnings per run. These are TVDb series IDs referenced in the upstream defaults (franchise, streaming, network, etc.) that no longer resolve on TVDb — likely removed, merged, or renumbered.

## Expected Behavior

Default collection files should reference valid TVDb IDs, or gracefully handle removed series without producing errors.

## Actual Behavior

Each run logs ~137 errors like:

```
TVDb Error: No Series found for TVDb ID: 463160 at https://www.thetvdb.com/dereferrer/series/463160
```

These are cosmetic (collections still work for valid IDs), but they clutter logs and trigger failure notifications via Notifiarr/Discord.

## Stale TVDb IDs

<details>
<summary>Full list of 137 stale TVDb series IDs (click to expand)</summary>

```
26021, 58039, 61103, 72278, 77279, 84482, 85036, 90261, 90271, 93581,
93711, 102221, 104053, 127181, 127701, 146086, 159884, 163754, 192551,
264026, 264082, 264368, 264462, 264467, 264647, 262885, 267966, 269025,
280819, 281919, 286008, 302343, 303045, 309457, 310520, 316192, 319337,
321553, 323843, 330422, 333695, 335276, 335985, 338052, 342585, 342833,
350900, 351532, 351566, 351735, 352287, 357127, 362588, 365826, 366054,
367922, 369458, 380612, 383314, 388118, 389293, 396282, 403568, 406646,
407415, 410424, 411222, 412403, 416651, 419338, 419705, 421873, 422168,
423203, 423239, 423603, 424260, 424677, 424933, 425011, 426391, 427794,
428197, 430152, 432557, 432598, 432967, 435922, 437845, 438305, 438518,
438737, 439718, 440144, 440145, 440357, 441702, 442833, 444327, 444823,
445455, 445530, 447313, 449073, 451792, 452341, 455512, 455857, 456879,
457049, 457080, 457165, 457992, 458966, 460780, 461362, 461498, 462238,
463160, 464881, 466661, 466926, 468867, 469302, 475020, 687821, 1907591,
1998714, 2043960, 2076134, 2111532, 2140968, 2207257, 5536816, 7918362,
8915358
```

</details>

## Environment

[...]

PR #3137, opened 2026-05-30
fix: skip deleted TMDb collections from defaults instead of failing

SturmB:
## Problem

The `franchise` default builds collections from a dynamic `tmdb_collection` type that scans every movie in the library and auto-discovers the TMDb collection ID cached on each (`meta.py`, `auto_type == "tmdb_collection"`). When TMDb **deletes** a collection — their policy now only permits collections for true movie sequels, so thematic groupings like *The Dollars Trilogy*, *The Cornetto Trilogy*, and *Miyazaki Movies* have been removed upstream — the stale collection ID lingers on the cached movie records.

The franchise default still generates a `tmdb_collection_details` builder for that ID, `validate_tmdb_ids` 404s, and the run aborts the collection with a **critical error plus a failure webhook** (Notifiarr / Discord):

```
TMDb Error: No valid TMDb IDs in [1664873]
```

The user never typed the ID and cannot act on it — it came from Kometa's own default scanning their library. This is the same class of problem as the stale-TVDb-ID fix in #3047 / #3129, applied to TMDb collections.

## Fix

- Add `modules.tmdb.NotFound` (subclass of `Failed`) for a TMDb resource that is gone upstream (HTTP 4xx). `get_collection` raises it instead of plain `Failed`. (tmdbapis' own `NotFound` is imported as `TMDbNotFound`.)
- `validate_tmdb_ids` logs per-ID `NotFound` at debug and, when **every** ID in the list is gone, raises `NotFound` rather than `Failed`. A mix that includes any non-`NotFound` failure still raises `Failed`.
- In the builder gather loop, catch `tmdb.NotFound`: if the collection came from a **Default** file (`self.metadata.type == "Default"`), raise `NonExisting` so it is reported as **Ignored** with no webhook — matching how Kometa already handles other non-actionable skips. `tmdb_collection_details` in **user-authored** files still raises, since those IDs are the user's responsibility.
- Add `tests/test_tmdb.py` covering `NotFound` subclassing, `get_collection` 4xx mapping, and `validate_tmdb_ids` all-deleted vs partial-failure vs valid behavior.

## Testing

`tests/test_tmdb.py` passes (5/5). Validated live against a real library: with the deleted IDs excluded from the franchise default, the run completes with zero errors and no failure webhooks; user-authored TMDb collection builders are unaffected.


chazlarson:
In future, please use the standard PR template.

PR #3152, opened 2026-06-12
fix: graceful handling of stale TMDb collection IDs and TVDb 5xx retry gap

badja-dev:
## What type of PR is this?

## Description

Two related error-handling gaps in the TMDb and TVDb modules:

**`modules/tmdb.py` - stale collection ID 404s give no actionable guidance**

When a TMDb collection ID returns 404 (collection removed or renumbered on TMDb), the error was raised as `Failed`, indistinguishable from any other fatal failure. This particularly affects users of the `franchise` default, where collection IDs come from Plex metadata - users never hardcode them and have no obvious path to resolution.

Fix: introduces `NotFound(Failed)` subclass, mirroring the existing pattern in `tvdb.py`. `get_collection`, `load_movie`, and `load_show` now raise `NotFound` for 404s. `validate_tmdb_ids` catches `NotFound` separately and emits type-aware guidance - for Collection IDs it specifically hints that the user can add the ID to `exclude` in `template_variables`; for Movie/Show IDs it gives a generic "verify it still exists" message. The `tmdbapis` `NotFound` import is renamed to `TMDbNotFound` to avoid collision.

**`modules/tvdb.py` - transient 5xx server errors were not retried**

TVDb 5xx responses were raised as `Failed`, which is excluded from tenacity's retry condition (`retry_if_not_exception_type(Failed)`). This meant a single flaky TVDb response would immediately abort the item rather than using the retry budget.

Fix: introduces `TVDbServerError(Exception)` - not a `Failed` subclass, so tenacity retries it. `get_request` raises `TVDbServerError` for 5xx. `TVDbObj.__init__` and `get_id_from_url` catch it and convert to `Failed` after the retry budget is exhausted.

Both changes are internal only - no config schema or user-facing behaviour changes.

Tested locally by running Kometa against a config that references the two stale collection IDs (1706679, 1701553). Previously these produced generic `Failed` errors with no guidance; after this change they produce the actionable `exclude` hint.

- [ ] No - `exclude` is already documented; our change just references it in an error message.

- [X] Yes - two entries added under # Bug Fixes.


antwanchild:
@badja-dev PRs need to be submitted against the nightly branch.

PR #3153, opened 2026-06-12
fix: resolve remaining <<vars>> in dynamic collection title_format

### LINKED ISSUE #1904: Defaults Collections with `<<` `>>` variables not picked up by `delete_collections` operation
### Version Number

nightly34

### What branch are you on?

nightly

### Describe the Bug

There is an issue with the `delete_collections` operation where collections which contain `<<``>>` variables are not being seen as configured.

In this example, PMM has the collection as `"Netflix Top <<limit>>` and the final name of the built collection is `Netflix Top 10`. 
When I then try to run a `delete_collections` operation the name is not matching because it is searching for `Netflix Top <<limit>>` as a string, which is not matched, so my actual `Netflix Top 10` appears to be unconfigured.

This is repeatable across most (but not all) collections with variables in the name.

image

image

image

```yml
libraries:
  Movies:
    collection_files:
      - pmm: streaming
      - pmm: content_rating_us
    operations:
      delete_collections:
        managed: true
        configured: true
        less: 99999
```

### Logs

_No response_

badja-dev:
**What type of PR is this?**

**Description:**

When Kometa builds dynamic collection titles, `title_format` substitution only handled `<<title>>` and `<<key_name>>` explicitly. Any other template variables were left as unresolved literals in the registered collection name - for example `<<limit>>`, which the streaming default carries as `limit: "500"` in its template defaults.

This caused `delete_collections: configured: true` to incorrectly treat those collections as unconfigured: the Plex collection title (fully resolved, e.g. `Netflix Top 500`) didn't match the registered name (still containing `<<limit>>`), so Kometa deleted collections it had built itself.

**Fix:** after the initial `title_format` substitution, two additional resolution passes run:

1. **`og_call` variables** - per-key values from the dynamic collection's `template_variables` (e.g. `<<key>>`, `<<style>>`)
2. **Template defaults** - variables defined in the `default:` block of the referenced template(s) (e.g. `<<limit>>` from the `streaming` template)

The second pass is guarded by `if "<<" in collection_title` so it's skipped entirely if the first pass already resolved everything.

**Tested locally** by reproducing the exact #1904 scenario: streaming default with `title_format: "<<key_name>> Top <<limit>>"` and `limit` deliberately not set in `template_variables` (forcing it to resolve from template defaults). Without the fix, the collection registers as `Netflix Top <<limit>>`. With the fix, it registers as `Netflix Top 500`.
image

**Related Issues:**
- Closes #1904

- [ ] No *(no user-facing config changes)*

- [X] Yes


antwanchild:
Will this handle some of the defaults like some of the awards collections that get deleted on every run even though they are technically configured?

badja-dev:
@antwanchild 
> Will this handle some of the defaults like some of the awards collections that get deleted on every run even though they are technically configured?

My fix does not affect award collections. I tested default: award/oscars against unpatched nightly - all titles register cleanly: Oscars Winners 2026, Oscars Best Picture Winners, etc. No unresolved <<vars>> anywhere in the collection titles.

This is expected - every award default uses title_format: "<Name> <<key_name>>" only, and <<key_name>> was already resolved by the original code. My fix only adds resolution for variables beyond <<key_name>> and <<title>> (e.g. <<limit>> from template defaults, <<key>> from per-key vars).

Could you please expand?

image

badja-dev:
@antwanchild 
I tested default: award/bafta with delete_collections: configured: false against a live Plex. BAFTA 2023, 2025, 2026 survived - they were correctly in self.library.collections because title_format: "BAFTA <<key_name>>" resolves cleanly with only <<key_name>>. This confirms my fix doesn't affect default award collections.
If you're seeing specific awards collections deleted on every run, I'd like to know: which default, which settings, and whether it's year-based collections (sync window sliding) or static ones like "BAFTA Best Films". The two deletion pathways (delete_collections via operations.py and sync: true via meta.py) behave differently, and the reproduction case would help narrow it down.
Is there an existing bug?

antwanchild:
It's been ongoing for awhile. But it's Cesar, SAF, Peoples Choice, they all get deleted on each run. Sounds like this falls under something else. And they are year based

PR #3154, opened 2026-06-12
Fix: configured collections incorrectly appear as unconfigured when run_order places operations before collections

### LINKED ISSUE #1968: Collections appear as Unconfigured when `run_order` has operations prior to collections
### Version Number

1.21

### What branch are you on?

nightly

### Describe the Bug

There is some inconsistent reporting on Unconfigured collections when `run_order` has operations happening prior to collections.

For example my YAML is:
```yaml
collections:
  Thriller:
    plex_all: true
```
and my run order is

When my run order is `Operations, Metadata, Collections, Overlays` I get this:
image

If I change my run order to `Collections, Metadata, Operations, Overlays` I get:
image

This impacts both the `show_unconfigured` setting and the `delete_collections` operation when `unconfigured: true`. It should be noted that even though the collection is shown as unconfigured, the collections are not deleted as part of the operation.

_No response_

### Logs

_No response_

badja-dev:
**What type of PR is this?**

**Description:**
When `run_order` places `operations` before `collections`, `library.collection_names` is still empty at the point operations checks it. `operations.py` uses `col.title not in self.library.collection_names` to decide whether a Plex collection is unconfigured — so every collection in the library is flagged as unconfigured, regardless of whether it is defined in a collection file.

`collection_names` is only populated as each collection is built during the collections run phase. If operations runs first, it never gets that chance.

**Fix:** before the `run_order` loop begins, iterate through all collection files and call `get_collections()` to pre-seed `library.collection_names` with each collection's resolved title. `get_collections()` is a lightweight YAML read with no Plex API calls. A duplicate guard prevents double-entries when the collections phase later appends the same names.

**Tested locally** by reproducing the #1968 scenario: a static collection (`BBC`) defined in a collection file, with `run_order: [operations, metadata, collections, overlays]`. Without the fix, `BBC` is reported as unconfigured when operations runs first. With the fix, it is correctly identified as configured regardless of run order.
image

**Related Issues:**
- Closes #1968

- [ ] No *(no user-facing config changes)*

- [X] Yes


YozoraXCII:
Tested and works as expected

Issue #3168, opened 2026-06-16
Dynamic Award Year Collections are deleted and recreated on every run


antwanchild:
### Version Number

2.3.1

### What branch are you on?

nightly

### Describe the Bug

Dynamic Year Collections in the defaults for award collections for SAG, Cesar, seem to get deleted at the start of every run. these are deleted whether I have delete_collections or not.

and when delete_collections is run, all my birthday collections, people's choice awards, spirit best feature, SIMKL, a couple letterbox collections (all of the above are defaults), I also have setup (using defaults) the birthday collections. these show as unconfigured and are deleted 

### Relevant Collection/Overlay/Playlist Definition

```yml

```

### Logs

[meta.log.zip](https://github.com/user-attachments/files/29012080/meta.log.zip)

_No response_

antwanchild:
[meta.log.zip](https://github.com/user-attachments/files/29115377/meta.log.zip)

<img width="868" height="146" alt="Image" src="https://github.com/user-attachments/assets/d613ae88-1169-4c17-8d1d-82ab29afd0e2" />

Mine actually may be a different issue than what Badja was fixing

YozoraXCII:
Closing, will use #3189 going forward

PR #3187, opened 2026-06-19
title_Format update for SAG and Cesar

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other 

## Description

existing synced collections are named Screen Actors Guild Awards [year] through [year]
Kometa is generating Screen Actors Guild 2026, Screen Actors Guild 2025, etc.
the debug line shows in_sync=False for every generated title
so nothing gets removed from sync
then the delete loop removes all the old collections 
The reason is the title template:
log says Title Format: Screen Actors Guild <<key_name>>
but the existing collection names include Awards
So this one is not a delete bug either. It’s a naming mismatch:
current generated title: Screen Actors Guild 2026
existing Plex title: Screen Actors Guild Awards 2026
To stop deletion, the generated title has to match the existing one exactly. The simplest fix is likely to change the dynamic collection title_format to include Awards

## Related Issues [optional]

- Related Issue #3168 
- Closes #

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

### DIFF
--- a/defaults/award/cesar.yml
+++ b/defaults/award/cesar.yml
@@ -37,7 +37,7 @@ dynamic_collections:
       event_id: ev0000157
       starting: latest-4
       ending: latest
-    title_format: Cesar <<key_name>>
+    title_format: César <<key_name>>
     template:
       - use_year_collections
       - imdb_award
--- a/defaults/award/sag.yml
+++ b/defaults/award/sag.yml
@@ -41,7 +41,7 @@ dynamic_collections:
       event_id: ev0000598
       starting: latest-4
       ending: latest
-    title_format: Screen Actors Guild <<key_name>>
+    title_format: Screen Actors Guild Awards <<key_name>>
     template:
       - use_year_collections
       - imdb_award


antwanchild:
When I ran locally, the collections were created on the first run, then when i ran a subsequent run, the collections SAG and Cesar were not deleted as in previous runs

bullmoose20 (commented):
Not sure if your fix would actually work or create duplicates for people that already have those collections

Because one now has accents while the other does not… for Cesar.

And one contains « awards » while others don’t

antwanchild:
> Not sure if your fix would actually work or create duplicates for people that already have those collections
> 
> Because one now has accents while the other does not… for Cesar.
> 
> And one contains « awards » while others don’t

and if they did, it would delete them when the operation is done and then they'd be recreated; but will not be deleted again.

This I tried this several times and set it, then ran with the current default, then ran with this fix.  Each time running with my pr, I no longer get the deleted and recreated collections like we have had with a couple of the dynamic awards collections

YozoraXCII:
> > Not sure if your fix would actually work or create duplicates for people that already have those collections
> > 
> > Because one now has accents while the other does not… for Cesar.
> > 
> > And one contains « awards » while others don’t
> 
> and if they did, it would delete them when the operation is done and then they'd be recreated; but will not be deleted again.
> 
> This I tried this several times and set it, then ran with the current default, then ran with this fix.  Each time running with my pr, I no longer get the deleted and recreated collections like we have had with a couple of the dynamic awards collections

Not everyone is using the operation though

YozoraXCII:
closing and raising as an issue to sort this properly as there's a bug

PR #3191, opened 2026-06-19
fix: honour title_format and name_format in dynamic collection sync

### LINKED ISSUE #3189: `title_format` not being respected for `dynamic_collections`
### Version Number

2.3.1-build53

### What branch are you on?

nightly

### Describe the Bug

There is an ongoing issue where `title_format` is not being respected when used with `dynamic_collections`, the end result is that when a dynamic collection has `sync: true` set, the collections are repeatedly deleted and recreated.

using the `sag` defaults file as an example:
```yaml
dynamic_collections:
 Screen Actors Guild Awards:
    type: imdb_awards
    sync: true
    data:
      event_id: ev0000598
      starting: latest-4
      ending: latest
    title_format: Screen Actors Guild <<key_name>>
```

based on `title_format`, the end result in Plex should be `Screen Actors Guild 2025` for example, and this can be seen in the Kometa summary that it has been created:

Image

However, when looking in Plex, the collection created is actually `Screen Actors Guild Awards 2025`:

Image

This creates a loop on the next run as the titles no longer match, so the collection gets deleted and recreated.

Log attached

```yml
libraries:
  Movies:
    collection_files:
      - default: sag
        template_variables:
          data:
            starting: latest-6
```

### Logs

https://gist.github.com/YozoraXCII/ae2ef1bf0432c74185a87991c102675e

badja-dev:
**What type of PR is this?**

**Description:**

Dynamic collections with an explicit `title_format` or `name_format` are silently overridden by `translation_key` when resolving the final Plex collection name. Because the name Kometa uses internally (`collection_title` in `meta.py`) never matches the name it writes to Plex (resolved by `builder.py` via `translation_key`), `sync: true` treats every collection as an orphan on every subsequent run - deleting and recreating it indefinitely.

Two affected cases:

- **`title_format` (e.g. SAG defaults):** `title_format: Screen Actors Guild <<key_name>>` resolves to `Screen Actors Guild 2025`, but `translation_key: sag_year` causes builder.py to name the Plex collection `Screen Actors Guild Awards 2025`. Sync deletes `Screen Actors Guild Awards 2025` as an orphan, then recreates it - on every run.
- **`name_format` (e.g. birthday collections):** `name_format: 🎂 Happy Birthday <<key_name>>` flows through template substitution in `apply_template` to name the Plex collection `🎂 Happy Birthday Tom Hanks`. But `meta.py` tracks the collection under the bare `title_format` default (`Tom Hanks`). The same orphan-delete loop occurs.

**Fix:** in `meta.py`'s dynamic collection expansion loop, the `collection_title` (sync key) is now derived from the explicitly-set format rather than always from `title_format`:

1. **`title_format`** - if it differs from the default, `collection_title` is computed from it and set as `col["name"]`, so `builder.py` uses it directly instead of falling back to the `translation_key` result.
2. **`name_format`** - checked in `og_call` first (for defaults YAMLs that declare it in their own `template_variables` block), then in `self.temp_vars` (for user-supplied `template_variables` that the defaults YAML doesn't forward). If present, used as the base for `collection_title` so the sync key matches what the builder will actually write to Plex.

Setting `col["name"]` in both cases also prevents `apply_template` from overriding the name during template expansion, since builder.py skips template-derived `name` attributes when `name` is already present in the collection data.

**Tested locally** against a `Movies-Kometa-Test` library running four sequential Kometa passes (two on stock nightly, two on the fix branch):

| Run | Image | Dynamic deletions | Notes |
|---|---|---|---|
| BUG_RUN1 | nightly | 25 | Birthday collections deleted on first run (wrong sync key) |
| BUG_RUN2 | nightly | 28 | SAG + birthday all deleted - loop confirmed |
| FIX_RUN1 | fix branch | 3 | Old wrong-named SAG collections cleaned up; birthday stabilised |
| FIX_RUN2 | fix branch | 0 | All 3 SAG year + 25 birthday collections Unchanged - fix confirmed |

**Related Issues:**
- Closes #3189

- [ ] No - this restores intended behaviour; no new options or behaviour changes to document

- [X] Yes


YozoraXCII:
CHANGELOG conflict - can't test until later today

PR #3192, opened 2026-06-19
feat: add hub_priority and auto_sort_hubs for Recommendation Hub ordering

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Adds two new attributes to control the order of Recommendation Hub rows on the Plex home screen and library Recommended tab, applied after all collections have been processed.

**`auto_sort_hubs`** (global / per-library setting) - sorts all collection hub rows for a library after the collections phase completes. Accepted values: `sort_title`, `sort_title.desc`, `alpha`, `alpha.desc`, `added`, `added.desc`, `random`. Only collections promoted to a hub (`visible_library`, `visible_home`, or `visible_shared`) are affected; built-in Plex section hubs (Recently Added, Top Unwatched, etc.) are not touched.

```yaml
libraries:
  Movies:
    settings:
      auto_sort_hubs: sort_title
```

**`hub_priority`** (per-collection attribute) - pins a collection's hub row to a specific position at the front of the sorted list. Collections with `hub_priority` are always sorted before those without, lowest number first, tiebroken by sort title. Requires the collection to be promoted to a hub.

```yaml
collections:
  My Important Collection:
    visible_home: true
    hub_priority: 1
```

Implementation: `sort_collection_hubs()` in `plex.py` fetches hub order via `GET /hubs/sections/{key}/manage` using raw XML (`_query()`). Hub elements use `identifier="custom.collection.{sectionKey}.{ratingKey}"` - the rating key is parsed from the identifier and the full identifier string is used in the move URL (`PUT /hubs/sections/{key}/manage/{identifier}/move`). `hub_priorities` is populated in `builder.py` during `update_details()` and passed to `sort_collection_hubs()` after all collection files have run.

Tested against a dedicated `Movies-Kometa-Test` library with 4 collections - 3 with `hub_priority` (1, 2, 3), 1 without:

| Run | Hub sort ran | Sort complete | Hubs sorted |
|---|---|---|---|
| Stock nightly (baseline) | No | - | - |
| Feature branch | Yes | Yes | 4 |

Visual order in Plex confirmed: Hub Test Alpha -> Hub Test Gamma -> Hub Test Beta -> Hub Test Unprioritised ✓

## Related Issues [optional]

- Supercedes #2911

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


YozoraXCII:
This is excellent, great work.

Some feedback points:

**WORKING:**
- `hub_priority` sets priority as expected
- duplicate priority falls back to sorting alphabetically
- if not all collections have `hub_priority`, the ones that do are sorted first and the rest fall back to alphabetical

**SUGGESTED CHANGES:**
- for points 2 and 3 above, perhaps the fallback should be whatever `auto_sort_hubs` order is set, and if none is set then fallback to alphabetical?

**NOT WORKING/NEEDS CONSIDERATION:**
- Worried about the `added` option and how that would be perceived - if the collection already exists in Kometa, it isn't "added" as part of the run. Maybe it's changed to "configured" and it's the order in which Kometa reads the collections from the config?

- Plex's "out of the box" collections seem to be ignored, this is probably by design as Kometa doesn't know they exist - do we need to cover this in docs, or is there a way to grab the list of Hub Recommendations and include them in `auto_sort_hubs`?

- Mixed results with the `auto_sort_hubs` values.

yaml used:
```
collections:
  Collection 1:
    sort_title: A
    visible_home: true
    plex_search:
      all:
        title: The
  Collection 2:
    sort_title: C
    visible_home: true
    plex_search:
      all:
        title: The
  Collection 3:
    sort_title: B
    visible_home: true
    plex_search:
      all:
        title: The
```
auto_sort_hubs values:

- `added` - works

- `added.desc` - DOES NOT WORK
    expected sort order: Collection 3, Collection 2, Collection 1
    actual sort order: Collection 1, Collection 2, Collection 3

- `sort_title` - DOES NOT WORK (in an odd way):
    expected sort order: Collection 1, Collection 3, Collection 2
    actual sort order: Collection 2, Collection 1, Collection 3

- `sort_title.desc` - DOES NOT WORK (in an odd way):
    expected sort order: Collection 3, Collection 2, Collection 1
    actual sort order: Collection 3, Collection 1, Collection 2

- `alpha` - works

- `alpha.desc` - works

- `random` - works

PR #3238, opened 2026-06-23
tweak: log line wrapping

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Python's `%-Ns` format pads strings to a minimum width but never truncates, so log messages longer than `screen_width` overflow the pipe-bordered log box.

This PR adds `textwrap`-based wrapping inside `MyLogger._log` for two cases:

**Bordered messages** (those starting with `| ` but not `|=`): content is extracted, wrapped at `screen_width - 4` characters, and each chunk is re-emitted as a correctly-padded `| chunk |` line.

**Plain messages** (those not starting with `|`): the message is wrapped at `screen_width - 4` characters and each chunk is re-emitted as a standalone line.

In both cases, continuation lines (second chunk onwards) are prefixed with two spaces so they are visually distinct from the first line. Separator divider lines (`|===...===|`) are excluded from wrapping as they are always constructed to the correct width.

**Testing:** 9 new unit tests added to `tests/test_logs.py` covering short messages (no wrapping), separator passthrough, plain long-line wrapping, bordered long-line wrapping, first-line/continuation indent behaviour, and a width constraint check for each wrapped output line.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


antwanchild (approved):
Looks good to me

YozoraXCII:
@badja-dev can you review the conflicts and then this can be merged

PR #3239, opened 2026-06-23
infra: introduce pyright with ratcheting baseline

chazlarson:
## What type of PR is this?

## Description

Replaces the existing `mypy` (informational-only) setup with `pyright` using a **ratcheting baseline** approach. Instead of demanding zero type errors on day one (impossible on a codebase this size — pyright reports 1223 errors against `modules/` + `kometa.py` today), this approach pins the current per-file error counts as a baseline and **fails CI on any PR that introduces new errors in a file**, while letting maintainers chip away at existing errors at their own pace.

### Why

`mypy` is in `dev-requirements.txt` but its CI job runs with `continue-on-error: true`, so it produces output nobody reads. Pyright with a ratcheting baseline:

- gives **actionable** signal (a PR that adds new errors fails CI, with a per-file breakdown of what changed)
- doesn't demand a giant up-front "fix all 1223 type errors" refactor
- naturally trends error count downward over time (any PR that *removes* errors implicitly tightens the baseline if the maintainer reruns `--update`)
- catches the `None | T` cluster (`reportOptionalMemberAccess`, `reportOptionalSubscript`, `reportOptionalCall`) which is where most of Kometa's silent crashes hide

### How the baseline works

- `.pyright-baseline.json` — the accepted per-file error count. Generated once by running `scripts/pyright_baseline.py --update`.
- `scripts/pyright_baseline.py --check` — runs pyright, compares per-file counts to the baseline, and fails if any file's count went up. Currently used by the `pyright` CI job.
- `.github/workflows/pyright.yml` — runs `scripts/pyright_baseline.py --check` on every PR.

To **fix** errors:
1. Fix them in code.
2. Run `pip install -r dev-requirements.txt` (gets `pyright==1.1.410` pinned).
3. Run `python scripts/pyright_baseline.py --update` to lower the baseline.
4. Commit the updated `.pyright-baseline.json` alongside your fix.

To **add a new file** that has zero errors:
- No action needed — it'll just naturally have a `0` entry next time the baseline regenerates.

To **silence a single line** (use sparingly):
```python
foo.bar  # pyright: ignore[reportOptionalMemberAccess]
```

To **opt a single file into stricter checking** (the long-term goal for every file):
```python
# pyright: strict
```

### What changes

- `dev-requirements.txt` — `mypy==2.1.0` → `pyright==1.1.410` (keeps the existing test deps from #3232)
- `pyproject.toml` — adds `[tool.pyright]` config block (replaces `[tool.mypy]`)
- `.github/workflows/pyright.yml` (new) — dedicated CI job that runs the baseline check
- `.pyright-baseline.json` (new) — per-file error counts captured in a clean Python 3.12 env (matching CI)
- `scripts/pyright_baseline.py` (new) — the check/update script
- `scripts/README.md` (new) — usage docs for the script, with a warning about regenerating in a clean venv to match CI's environment exactly
- `.github/.wordlist.txt` — adds `pyright`, `pyright's`, `stdlib`, `tmp` (the docs reference each)

### Verification

- Baseline runs cleanly: `1223 errors baselined, 1223 current, +0 delta`
- 660/660 pytest still passes (no behavioral changes)
- Baseline was regenerated in a clean Python 3.12 venv with `pip install -r requirements.txt` to match CI exactly — different Python versions report slightly different error counts (3.14 in my dev env reported 1218; 3.12 in CI reports 1223). The check script normalizes by per-file count, not total, so this doesn't matter for the gate, but it does matter for the baseline file itself.

### Future cleanup

After this merges, every `reportOptionalMemberAccess` fix becomes a small, mechanical PR:

```python
# Before
result = some_fn_returning_optional()
result.attr  # pyright complains: result might be None

# After
result = some_fn_returning_optional()
if result is None:
    raise Failed("descriptive message")
result.attr
```

The `reportOptionalMemberAccess` cluster alone is ~400 of the 1223 errors. Mostly low-risk, drives baseline down quickly.

### What's NOT in this PR

- No code fixes — that's deliberately the next round of PRs. Mixing infra + fixes makes review impossible.
- No removal of `# type: ignore` comments — those stay for now, can be cleaned up over time as pyright covers the cases they were silencing.
- No `--strict` mode on any file yet.

## Related Issues [optional]

- Related Issue #
- Closes #

- [X] Yes

- [X] Not Applicable

- [X] Yes


chazlarson:
Pushed one more commit (`86e09fcb`): `ci: remove orphan mypy step from test.yml`.

Caught this during a risk-assessment pass on the PR. The original infra commit (`23878909`) swapped mypy out of `dev-requirements.txt` but missed this step in the lint job:

```yaml
- name: mypy
  run: mypy modules/
  continue-on-error: true
```

After this PR merges, the step would shell out to a binary that's no longer installed, fail with `command not found`, and be silently swallowed by `continue-on-error: true` — leaving a yellow on every PR's lint job going forward. Cleaner to remove it here.

Also updated the workflow header comment so the listed checks (`black, isort, flake8, bandit`) match reality.

Notes:
- YAML re-parses cleanly; lint job steps are now `checkout -> Set up Python -> Cache pip -> Install deps -> Black check -> isort check -> flake8 -> bandit`.
- `lint.yml` also references mypy but triggers on `disabled_for_now` (a fake branch), so it's dormant. Left it alone — out of scope, will get addressed if/when that workflow is revived.
- Rebased #3258 (poster fixes) on top of this so it stays in sync.

PR #3271, opened 2026-06-25
Fix/relative paths config dir

### LINKED ISSUE #3092: Relative paths in config resolve against CWD instead of config file directory when using custom config path

badja-dev:
## What type of PR is this?

- [X] Documentation Update
- [ ] Other

## Description

When Kometa is launched with a custom config path via `-c` (or the `KOMETA_CONFIG` env var), relative paths defined inside that config were resolved against the process working directory instead of the directory containing the config file. This meant `- file: my_collections.yml` would silently fail to load if Kometa was not started from the same directory as the config.

The affected keys were `collection_files`, `metadata_files`, `overlay_files`, `playlist_files`, `image_files`, `folder`, and `asset_directory`.

**Root cause:** `util.load_files()` had no concept of a base directory. The three code paths that check whether a path exists (`folder`, `asset_directory`, and plain string entries) all called `os.path.isdir()` / `os.path.exists()` directly on the raw string from the YAML, which Python resolves relative to CWD. Dict-form `file:` entries were stored as-is and then hit the same problem later in `meta.py` via `os.path.abspath()`.

**Fix:**

- Added a `config_dir=None` parameter to `util.load_files()` with a `_resolve()` helper that anchors relative paths to `config_dir` when provided and leaves absolute paths untouched.
- Applied `_resolve()` to all three affected code paths (`folder`, `asset_directory`, plain string entries), and also to dict-form `File` entries before they are stored in the returned list, so downstream code always receives absolute paths.
- Passed `config_dir=self.default_dir` at all six `util.load_files()` call sites in `config.py` (`playlist_files`, `collection_files` x2, `metadata_files`, `overlay_files`, `image_files`).
- Added a defensive fix in `meta.py` `load_file()` to anchor any remaining relative `File` paths to `self.config.default_dir` before calling `os.path.abspath()`.

The fix is non-breaking: users running Kometa from the same directory as their config see identical behaviour, since `config_dir` equals CWD in that case.

Tests covering all three path types (plain string, dict `file:`, `folder:`, `asset_directory:`) with relative, absolute, missing, and CWD-fallback cases were added to `tests/test_util.py`. All 24 tests pass.

## Related Issues

- Closes #3092

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


chazlarson:
I don't understand why this needs to be "fixed" when the solution is a simple "don't do that".

What is the use case where running Kometa from some random directory is required?

badja-dev:
> What is the use case where running Kometa from some random directory is required?

Funnily enough I thought the same, but the issue didn't get closed, so I assumed you were explaining the situation and requesting a fix 😂

chazlarson:
Oh, I just meant to clarify that the stated issue is not accurate.  I run nearly every run with a custom config file with no issues.  It's not custom config files that are the problem, it's that custom config file being in some other directory.  If you make all relative paths relative to whatever directory the config happens to be in the defaults stop working unless they are all duplicated or dealt with specially.

Just seems like a can of torpedoes for not much gain.

YozoraXCII:
So what are we doing with this PR and the related issue? I'm confused - are we merging this, or closing both the PR and related issue?

chazlarson:
To me this feels like "stop hitting yourself".

YozoraXCII:
so close both?

YozoraXCII:
Closing

PR #3317, opened 2026-07-03
fix: security hardening and cache/request performance improvements

chazlarson:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Security hardening and performance improvements from a codebase review, all non-breaking.

**Security**

- `modules/request.py`: every outbound HTTP request now sends a 30-second per-socket timeout (`DEFAULT_TIMEOUT`). Previously no `Requests.get`/`post` call had a timeout, so a single stalled external server (TMDb, an image host, etc.) would hang the entire run forever — tenacity never fires because a silent stall raises nothing.
- `modules/cache.py`: dynamic table/column names interpolated into SQL are validated by a new `sql_identifier()` guard at every method that accepts them. All current callers pass internal identifiers, so this hardens against future misuse rather than fixing an exploitable path. The reviewed bandit B608 sites (and the scheme/host-guarded B310 `urlopen` in `letterboxd.py`) now carry `# nosec` annotations — `bandit -r modules/ kometa.py -ll` reports zero findings.
- `modules/request.py`: a scoped `verify_ssl: false` session (e.g. per-library Plex) no longer disables `InsecureRequestWarning` process-wide; only the config-level global SSL opt-out does.
- `modules/letterboxd.py`: short `boxd.it` URLs must be http/https; other schemes fail with a clear error before any network call.
- `modules/logs.py`: registered secrets are also redacted in their URL-encoded (`quote`/`quote_plus`) forms, so tokens embedded in logged query strings no longer slip past redaction.

**Performance**

- `modules/cache.py`: one shared SQLite connection (WAL journal mode, `row_factory` set once) replaces per-call `sqlite3.connect` at all ~60 sites — the old code opened a fresh connection per query and never closed it. The overlay path makes 3–5 cache calls per library item, so large runs opened tens of thousands of connections. Per-block commit semantics are unchanged.
- `modules/request.py`: retries switch from a fixed 10-second wait (up to 50s of sleeping per failing URL) to exponential backoff capped at 10 seconds (~25s worst case, much less for transient blips).
- `modules/request.py`: `get_stream` throttles download-progress logging to ~4 updates/second plus a final 100% line, instead of one `logger.ghost` per 8 KB chunk.

**Testing**

- 22 new tests across `tests/test_request.py`, `tests/test_cache.py`, `tests/test_logs.py`, `tests/test_letterboxd.py`, written test-first (confirmed failing before implementation). Full suite: 749 passed.
- black / isort / flake8 / pyspelling pre-commit hooks pass; bandit medium+ findings: 17 → 0.
- Smoke-tested the real `Cache` (WAL mode, image-map round trips, identifier guard, cross-connection reads) and full `Requests` init.

## Related Issues [optional]

- Related Issue #
- Closes #

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/cache.py
badja-dev, line 1: Minor: every other f-string table-name in this file now goes through `sql_identifier()` (e.g. `query_image_map`, `update_overlay_state`, `query_arr_adds`, etc.), but `get_image_table_name()` still builds its `CREATE TABLE IF NOT EXISTS {table_name}...` / `{table_name}_backgrounds` / `{table_name}_logos` / `{table_name}_overlays` / `{table_name}_square_arts` / `{table_name}_overlay_state` / `{table_name}_overlay_images` statements directly, without the guard. Not exploitable today - `table_name` is always `image_map_<autoincrement int>` from `image_maps.key`, never user input - but it breaks the "every interpolated identifier passes through `sql_identifier()`" goal here
Worth a small change imo <3

YozoraXCII:
Did a few runs back-to-back with no issues, merging

PR #3354, opened 2026-07-11
Add YamTrack connector

YozoraXCII:
## What type of PR is this?

- [X] Documentation Update
- [ ] Other

## Description

Adds YamTrack connector support and a new `yamtrack_list` collection builder.

This PR allows users to configure a YamTrack instance in `config.yml`, then build Kometa collections from YamTrack list pages. Kometa logs in with the configured YamTrack credentials, loads the list page, and extracts TMDb IDs from YamTrack item links.

Example connector config:
```yaml
yamtrack:
  url: https://yamtrack.example.com
  username: dummy_user
  password: dummy_password
```

Example collection:
```yaml
collections:
  YamTrack Watchlist:
    yamtrack_list: https://yamtrack.example.com/list/1
    sync_mode: sync
```

The PR also adds `yamtrack_list_details`, which uses the first YamTrack list's description as the collection summary.
```yaml
collections:
  YamTrack Watchlist:
    yamtrack_list_details:
      - https://yamtrack.example.com/list/1
    sync_mode: sync
```

Included updates:
- New `modules/yamtrack.py` connector.
- Builder registration and summary handling.
- Config and collection JSON schema updates.
- YamTrack config docs.
- YamTrack builder docs.
- Focused unit tests for list parsing, description parsing, and login payload handling.

## Related Issues [optional]

- Related Issue #
- Closes #

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No


modules/yamtrack.py
```diff
@@ -0,0 +1,256 @@
...
+        if not ids and not mal_ids:
+            raise Failed("YamTrack Error: No IDs found in tracked items")
+        return ids, mal_ids
+
+    @staticmethod
+    def _default_statuses(enabled):
+        return {status: enabled for status in tracked_statuses}
+
+    @staticmethod
+    def _tracked_type_applies(media_type, is_movie):
```
chazlarson, line 209: `anime` is treated as cross-library here (`cross_library_tracked_types = ["anime"]` + `_tracked_type_applies()`), so `validate_tracked()` accepts anime for both movie and show libraries.

The problem is that downstream in `builder.py`, anime results are converted with `myanimelist_to_ids(mal_ids, self.library)`, which resolves against the current library only. So this code allows a config shape that the later conversion path can’t fully honor for single-type libraries.

Could we either:
- restrict `anime` to playlist mode only, or
- keep it enabled here but make the later conversion/media-type filtering explicit?

modules/builder.py
```diff
@@ -3512,6 +3532,13 @@ def gather_ids(self, method, value):
             ids = self.config.TMDb.get_tmdb_ids(method, value, self.library.is_movie, self.tmdb_region)
         elif "trakt" in method:
             ids = self.config.Trakt.get_trakt_ids(method, value, self.library.is_movie)
+        elif "yamtrack" in method:
+            if method == "yamtrack_tracked":
+                ids, mal_ids = self.config.YamTrack.get_tracked_ids(value, self.library.is_movie if not self.playlist else None)
```
chazlarson, line 3537: I think this is the blocking part of the anime support.

`get_tracked_ids()` can return MAL IDs from the YamTrack anime page, but `myanimelist_to_ids(mal_ids, self.library)` resolves against the current library only. In a movie library, anime entries that only map to shows will be missed; in a show library, anime entries that only map to movies will be missed.

That doesn’t match the current docs for `yamtrack_tracked`, which say both anime shows and anime movies can come from the same anime page.

I think this needs either:
1. playlist-only anime support, or
2. media-type-aware filtering/conversion so behavior matches the docs.

docs/files/builders/yamtrack/tracked.md
```diff
@@ -0,0 +1,57 @@
...
+
+In this example, Kometa adds only items marked `Planning` or `In Progress`.
+
+### Attributes
+
+| Attribute     | Description                             | Values         |
+|:--------------|:----------------------------------------|:---------------|
+| `movies`      | Movie statuses to include. Optional.    | Status block   |
+| `tv_shows`    | TV show statuses to include. Optional.  | Status block   |
+| `anime`       | Anime statuses to include. Optional. MyAnimeList IDs are resolved through Kometa's anime mappings, so both anime shows and anime movies can be found from the same YamTrack anime page. | Status block |
```
chazlarson: This looks a little too strong for the current implementation.

The docs say: “both anime shows and anime movies can be found from the same YamTrack anime page,” but the implementation converts MAL IDs against the current library (`myanimelist_to_ids(..., self.library)`), so a normal movie/show collection can only resolve what maps into that library type.

If anime is intended to be playlist-only, I’d document that explicitly. Otherwise this sentence should be narrowed so it doesn’t imply full cross-type support in single-type libraries.

docs/config/yamtrack.md
```diff
@@ -0,0 +1,34 @@
+# YamTrack
+
+Configuring [YamTrack](https://github.com/FuzzyGrim/YamTrack) is optional but is required for YamTrack list builders to function.
```
chazlarson: Small docs nit: this now supports more than just YamTrack list builders (`yamtrack_tracked` was added too), so “required for YamTrack list builders to function” should probably become “required for YamTrack builders to function”.

chazlarson (dismissed):
This review was done by Codex Code Review:

Nice feature overall — the YamTrack integration is cleanly scoped, tests are solid, and the builder/config/docs wiring is mostly in good shape.

However, I think there’s a blocking issue with yamtrack_tracked and the new anime support:

• validate_tracked() allows anime for movie/show libraries
• get_tracked_ids() returns MAL IDs for anime entries
• those MAL IDs are then converted with self.config.Convert.myanimelist_to_ids(mal_ids, self.library)

That conversion resolves against the current library only, so in a movie library any anime entries that map to shows will be dropped/missed, and vice versa in a show library. But the docs currently promise that “both anime shows and anime movies can be found from the same YamTrack anime page,” which overstates what a single-type library can actually do.

I think this should either:

• restrict anime to playlist mode only, or
• add media-type-aware conversion/filtering so anime resolves correctly for the target library type.

As written, I’d treat that as blocking because the feature behavior and docs don’t line up yet.

PR #3355, opened 2026-07-12
performance: batch per-item Plex writes and cut redundant network/cache calls on large runs

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

This PR batches per-item Plex writes and removes several redundant network/cache calls that were costing measurable time on large collection and library runs. The issues were found by profiling a full from-scratch run against a config exercising every collection, overlay, and operations feature at once, then validated with before/after timing comparisons on both that config and a copy of a real production config.

What changed:

- Batch the overlay "Overlay" label add, label/genre sync, label removal cleanup, mass rating sync, and title-parentheses cleanup into single batched Plex multi-edit calls instead of one editTags/editField/editTitle HTTP call per item.
- Dedupe forced Plex item reloads inside check_filters when multiple genre/label/collection filters are checked back to back on the same item.
- Early exit the check_for_var nested variable resolution loop once a pass makes no further substitutions, instead of always running a fixed number of passes.
- Buffer log handler flushes (every 25 records, or immediately on a WARNING/ERROR record) instead of flushing after every single line.
- Memoize cache.py ID-map lookups (query_tmdb_to_tvdb_map and siblings) in-process for the life of a run.
- Reuse a freshly downloaded local overlay image instead of re-fetching it from GitHub on every reference within a run.
- Memoize Requests.get_image() by URL for the run, so the same image URL is never fetched twice.
- Use HEAD instead of GET for image URL validation-only checks, avoiding downloading and discarding the full image body just to confirm a URL is valid.
- Use ruamel.yaml's safe loader for YAML loads that are never written back out (translation files, git/url/repo sourced default files, IMDb award-event data).
- Fix a real crash in the existing label-removal path (a list was passed where a set was expected).
- Fix the pyright CI baseline comparing file paths inconsistently across platforms (OS-native separators broke cross-platform comparison).

Testing: full pytest suite passes (869 tests), black/isort/flake8 clean, pyright baseline delta is zero. Each change was validated on live Plex data with before/after timing comparisons - the batched label/genre/rating/title writes cut per-item Plex write time substantially on the affected paths, cache memoization measurably cut ID lookup time, and log buffering removed real overhead on long runs. Behaviour is unchanged throughout - every change only reduces redundant work, verified by comparing run output before and after.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


modules/overlays.py
```diff
@@ -74,6 +74,8 @@ def run_overlays(self):
             logger.info("")
 
             total_keys = len(key_to_overlays)
+            # Items freshly composed this run - their "Overlay" label add is batched once after the loop.
```
YozoraXCII: Does this create an issue if a run dies during the Overlay process?

YozoraXCII: > Does this create an issue if a run dies during the Overlay process?

If yes, would it perhaps be safer to have a counter for the number of overlays and when it reaches `plex_bulk_edit_batch_size` it then applies, then resets the counter? So it does batches as it processes?

badja-dev: Labels were only flushed once at the end of the whole loop, so a mid-run crash would leave already-uploaded posters without their "Overlay" label (not data loss, just wasted resource next run

Now flushes every plex_bulk_edit_batch_size items instead of only at the end, same as the rest of the batched writes

Added `TestBatchAddLabel` for testing

badja-dev:
Superseded by #3382 and #3383, both merged into nightly. Closing

PR #3382, opened 2026-07-18
perf: add `--timings`/`KOMETA_TIMINGS` Runtime Flag for Timing Instrumentation (part 1)

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Adds an optional `--timings`/`--timing` runtime flag (`KOMETA_TIMINGS`/`KOMETA_TIMING` env var) that instruments network calls, cache lookups, and major per-collection/per-library phases with lightweight timing data. When enabled, a summary prints to the log at the end of the run, and a full per-source/per-library breakdown is exported as JSON and CSV files to the logs directory. Disabled by default, with negligible overhead when off.

Also adds an optional `KOMETA_PROFILE=pyinstrument` environment variable for deeper single-run profiling, writing an HTML report to the logs directory. This is aimed at maintainers diagnosing a specific slow run and requires the `pyinstrument` package (a dev dependency, not part of a normal Kometa install).

Both flags are wired into `kometa.py`'s existing `arg_data`/`run_args` system, the same one every other runtime flag (`--debug`, `--trace`, `--log-requests`, etc.) goes through, so the CLI flag, env var naming, and truthy-value parsing (`t`/`true`) are all consistent with the rest of Kometa.

Tested: full pytest suite passes (817 tests), black/isort/flake8 clean, pyright baseline delta is zero. This is purely additive instrumentation - disabled by default, and every entry point is a guarded early return when off, so there is no behavioural change to any existing run.

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


YozoraXCII:
For pyinstrument, how will that work in a docker environment when it's not part of the standard requirements?

badja-dev:
@YozoraXCII 
> For pyinstrument, how will that work in a docker environment when it's not part of the standard requirements?

So there are two options:
1. as is - pyinstrument is missing, users can add `KOMETA_TIMINGS ` or `--timings`
2. add `pyinstrument` to `requirements.txt` so the already-existing `KOMETA_PROFILE=pyinstrument` flag actually works in Docker instead of silently no-opping, it's a second, heavier profiler (more overhead, more detail) than `KOMETA_TIMINGS`, which is currently only in the testing branch for me to test timings with

PR #3385, opened 2026-07-20
Standardize Plex API retry handling

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Standardizes the generic Plex API retry policy across `modules/plex.py` so terminal Plex/Kometa failures are not retried and exhausted transient retries do not become opaque `tenacity.RetryError` exceptions.

The original failure was observed while reordering a collection. `moveItem()` caught a permanent Plex `BadRequest` and converted it to Kometa's `Failed`, but its Tenacity predicate still considered `Failed` retriable. Kometa waited through six attempts and then received `RetryError`, bypassing the collection sorter's existing `except Failed` handling.

Review identified the same repeated retry policy on the other Plex API wrappers. This PR now fixes the policy once and applies it consistently instead of limiting the correction to `moveItem()`.

### Shared retry policy

| Exception/result | Previous behavior | New behavior |
|:--|:--|:--|
| Plex `BadRequest` / HTTP 400 | Terminal only when allowed to escape directly; wrappers that converted it to `Failed` could retry six times | Always terminal |
| Plex `NotFound` / HTTP 404 | Terminal | Terminal |
| Plex `Unauthorized` / HTTP 401 | Terminal | Terminal |
| Kometa `Failed` | Could be retried by the repeated three-exception policy | Always terminal |
| Retriable transport/client exception | Up to six attempts | Up to six attempts |
| Exhausted retriable exception | Wrapped in `RetryError` | Original exception is re-raised with `reraise=True` |

The policy is defined once as `PLEX_RETRY` and reused by all 23 generic Plex wrappers. The specialized `saveMultiEdits()` timeout retry remains unchanged because it has its own attempt count, wait schedule, batch-state restoration, logging, and exhaustion callback.

### Affected wrapper categories

| Category | Examples |
|:--|:--|
| Search and fetch | `search`, `exact_search`, `fetchItem`, `fetchItems`, `get_tags` |
| Collections and playlists | `create_playlist`, `moveItem`, `query_collection`, `collection_mode_query`, `collection_order_query` |
| Generic calls and edits | `query`, `query_data`, `tag_edit`, `edit_query`, `_query` |
| Item loading | `item_labels`, `reload`, `get_actor_id` |
| Artwork | `_upload_image`, `upload_poster`, `upload_background`, `upload_logo`, `upload_square_art` |

### Collection move improvement

`moveItem()` continues to provide additional context specific to the originally reported failure.

| Behavior | Before | After |
|:--|:--|:--|
| Permanent Plex move failure | Retried six times | Fails immediately |
| Collection result | `Unknown Error: RetryError` | Existing collection-sorter `Failed` handling runs |
| Error message | Generic `Move Failed` | Includes the item title and original Plex response |
| Permanent failure delay | Approximately 50 seconds | No retry delay |

### Testing

| Validation | Result |
|:--|:--|
| Focused Plex tests | 30 passed |
| Full pytest suite | 795 passed |
| Black | Passed |

Regression coverage confirms:

- a known Plex collection-move `BadRequest` becomes `Failed` after one call;
- a generic callable's `Failed` exception remains terminal;
- a `BadRequest` wrapped by `reload()` remains terminal and is called once;
- a transient connection failure is attempted six times and then re-raises the original `ConnectionError`, not `RetryError`;
- the specialized multi-edit timeout tests continue to pass unchanged.

## Related Issues [optional]

None.

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [x] Yes
- [ ] No


YozoraXCII:
Probably one for @chazlarson or @badja-dev

badja-dev (commented):
This only fixes `moveItem` - other methods in `plex.py` with the same `retry_if_not_exception_type((BadRequest, NotFound, Unauthorized))` are not touched - and there's a lot of them.
Worth fixing at once?

antwanchild:
Expanded in ea21a1b8 as suggested. The repeated policy is now a shared `PLEX_RETRY` decorator applied to all 23 generic Plex wrappers: `BadRequest`, `NotFound`, `Unauthorized`, and `Failed` are terminal, while exhausted transient retries use `reraise=True` and surface the underlying exception instead of `RetryError`. The specialized `saveMultiEdits` retry remains unchanged. Added representative terminal/wrapped/transient-exhaustion tests; focused Plex tests: 30 passed, full suite: 795 passed.

YozoraXCII:
> Expanded in [ea21a1b](https://github.com/Kometa-Team/Kometa/commit/ea21a1b820c4bccd4b4704bfac10b93611abe5bc) as suggested. The repeated policy is now a shared `PLEX_RETRY` decorator applied to all 23 generic Plex wrappers: `BadRequest`, `NotFound`, `Unauthorized`, and `Failed` are terminal, while exhausted transient retries use `reraise=True` and surface the underlying exception instead of `RetryError`. The specialized `saveMultiEdits` retry remains unchanged. Added representative terminal/wrapped/transient-exhaustion tests; focused Plex tests: 30 passed, full suite: 795 passed.

@badja-dev happy with this?

badja-dev (approved):
Before this PR, two methods, `query_collection` and `get_actor_id`, were set up a bit differently than the rest of `plex.py`. Their retry logic only skipped retrying on Kometa's own `Failed` exception. That meant if Plex responded with something like a 400 Bad Request, these two would actually retry it six times before giving up, instead of failing fast like most of the other methods already did

Now that everything's been standardised under the shared `PLEX_RETRY` policy, that inconsistency goes away, which is good. But it does change how these two methods behave

Neither `query_collection` nor `get_actor_id` catches Plex's exceptions and turns them into `Failed` the way `moveItem` or `reload` do. So now, if Plex throws a `BadRequest`, it just passes straight through as-is. No retry, and no conversion into `Failed`

antwanchild:
Addressed in aee97bc5. `query_collection` and `get_actor_id` now catch `BadRequest`, `NotFound`, and `Unauthorized` at the method boundary and convert them to contextual Kometa `Failed` errors. Because `PLEX_RETRY` treats `Failed` as terminal, both methods still fail fast with a single Plex call, but no longer leak raw Plex exceptions. Added focused tests for both methods that assert the contextual error and exactly one underlying API call.

YozoraXCII:
@antwanchild can you update the wordlist file please
Beetlejuice
beetlejuice
filmography

PR #3390, opened 2026-07-21
Fix TMDb episode ratings for alternate Plex ordering

### LINKED ISSUE #3389: TMDb Episode Rating Positional Lookup Bug
### Version Number

2.4.4-build69

### What branch are you on?

nightly [corrected]

## Summary

Kometa's `mass_episode_audience_rating_update: tmdb` operation fails for split-cour anime (and any show where Plex/TVDB season numbering differs from TMDb's structure), even when Plex already has the correct TMDb episode ID stored in its metadata.

## Root Cause

Kometa constructs a positional API call using the show's TMDb ID + season number + episode number (e.g. `GET /tv/209867/season/2/episode/8`). This returns 404 for shows like Frieren: Beyond Journey's End where TMDb combines all episodes into a single season, while TVDB/Plex splits them into Season 1 and Season 2.

However, Plex already stores the individual TMDb episode GUID for each episode (e.g. `<Guid id="tmdb://6855841"/>`). Kometa ignores this GUID entirely and falls back to the positional lookup, which fails.

## Evidence

Frieren S02E08 XML shows:
- `<Guid id="tmdb://6855841"/>` — correct TMDb episode ID stored in Plex
- `<Rating image="themoviedb://image.rating" value="9.5" type="audience"/>` — correct rating fetched by Plex's metadata agent via episode ID

Yet Kometa logs: `TMDb Error: No Episode found for TMDb ID 209867 Season 2 Episode 8: (404 [Not Found])`

## Impact

- Errors logged on every Kometa run for affected shows (Frieren, Re:ZERO, Solo Leveling, Dan Da Dan, and others)
- Ratings display correctly in Plex because they are locked from Plex's own successful fetch
- Kometa cannot refresh/update those ratings going forward

## Proposed Fix

When processing episode-level TMDb rating updates, Kometa should first check whether the episode has a `tmdb://` GUID stored in Plex metadata and use that for a direct episode ID lookup (`GET /tv/episode/{id}`) before falling back to the positional lookup.

```yml

```

### Logs

_No response_

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Fixes TMDb episode rating resolution when Plex's selected episode ordering does not match TMDb's canonical season structure. This is commonly visible with split-cour anime, but it can affect any show using TVDb aired/DVD/absolute ordering or another numbering scheme that differs from TMDb.

Previously, Kometa ignored Plex's episode-level `tmdb://` GUID and always requested TMDb using the show's ID plus Plex's displayed season and episode numbers. For the example reported in #3389, Plex identifies Frieren S02E08 with `tmdb://6855841`, but Kometa requested `/tv/209867/season/2/episode/8`; TMDb represents those episodes under a different season structure, so the positional request returned 404.

### Behavior change

| Stage | Previous behavior | New behavior |
|:--|:--|:--|
| Episode identity | Show ID + Plex season/episode position only | Plex episode-level `tmdb://` GUID first |
| Alternate ordering | Could return 404 or resolve the wrong positional episode | Resolves the matching TMDb episode by its unique episode ID |
| Missing or stale GUID | Positional lookup | Positional lookup remains the fallback |
| Mass episode ratings | Positional TMDb lookup | Shared GUID-aware resolver |
| Overlay TMDb ratings | Positional TMDb lookup | Shared GUID-aware resolver |

Kometa does not modify Plex ordering, season placement, titles, or episode numbers. It only changes how the matching TMDb episode is identified when fetching its rating.

### Resolution flow

| Step | Action |
|:--|:--|
| 1 | Read the reloaded Plex episode's external GUIDs and extract a valid `tmdb://<episode_id>` value. |
| 2 | Query the existing TMDb episode cache by direct episode ID and language, and verify that the cached row belongs to the requested show. |
| 3 | On a cold or expired lookup, fetch each TMDb season once and build an episode-ID map from the returned episodes. |
| 4 | Bulk-store those episodes so subsequent direct-ID lookups are warm. |
| 5 | If no usable GUID or matching episode exists, use the existing show/season/episode positional lookup. |

Malformed GUID entries are skipped individually, so they cannot hide a later valid TMDb GUID.

The cold-map boundary also normalizes TMDb client failures and exhausted Tenacity retries to Kometa's `Failed` exception. This lets the shared Plex resolver attempt its positional fallback instead of allowing an unhandled `TMDbException` or `RetryError` to escape while loading the show or one of its seasons.

### Cache changes

This does **not** introduce or replace an episode cache table. It extends the existing `tmdb_episode_data2` table so the same canonical episode row can be queried by either identity.

| Cache detail | Before | After |
|:--|:--|:--|
| Existing table | `tmdb_episode_data2` | Same table |
| Positional identity | `tmdb_id`, `season_number`, `episode_number`, `language` | Unchanged |
| Direct identity | Not stored | Nullable `episode_id` column |
| Direct lookup index | None | Unique partial index on `episode_id, language` when `episode_id` is not null |
| Expiration | Existing episode-cache expiration | Same expiration applied to both lookup paths |
| Legacy rows | Positional data | Remain valid with `episode_id = NULL` until refreshed |
| Upgrade behavior | N/A | In-place `ALTER TABLE`; no table deletion or recreation |

When Kometa fetches an episode positionally, it now stores the returned episode's direct TMDb ID. When a season map is required, every returned episode is bulk-written with its show ID, canonical TMDb position, direct episode ID, language, rating, artwork, and existing metadata fields.

Season responses do not always include the richer IMDb and TVDb identifiers returned by a full episode-details request. Bulk season writes therefore preserve any non-null IMDb/TVDb IDs already present in the canonical cache row instead of replacing them with null values.

For example, Plex may display an episode as S02E08 while the cached canonical TMDb row records a different season/episode position. Both point to the same row through its unique TMDb episode ID, so Plex ordering no longer needs to match TMDb ordering.

### Request efficiency

| Scenario | Previous behavior | New behavior |
|:--|:--|:--|
| Cold positional run | Up to one details request per episode | Approximately one request per TMDb season when a GUID map is needed |
| Remaining episodes in the same cold run | Additional episode requests | Reuse the season map and bulk-populated cache |
| Warm run | Zero requests for warm positional rows | Zero requests for warm direct-ID rows |
| No episode GUID | Existing positional cache/lookup | Unchanged |

This provides GUID-first correctness without giving up the existing zero-request warm-cache behavior.

### Compatibility and fallback

| Condition | Result |
|:--|:--|
| Valid Plex TMDb episode GUID | Resolve through direct episode ID |
| No episode-level TMDb GUID | Use the existing positional lookup |
| Malformed TMDb GUID | Ignore it and continue checking other GUIDs |
| Stale, unmatched, or cross-show TMDb GUID | Reject the direct result and use the existing positional lookup |
| Existing cache database | Add the nullable column and index in place |
| Existing positional cache row | Continue using it; populate `episode_id` on refresh |

### Testing

| Validation | Result |
|:--|:--|
| Focused TMDb, Plex resolver, and real SQLite cache tests | 127 passed |
| Full pytest suite | 808 passed |
| Black | Passed |

Regression coverage includes alternate numbering, GUID-first selection, per-show map reuse, missing/stale/malformed/cross-show GUID fallback, show/season TMDb retry failures, warm persistent reads, cold bulk writes, preservation of richer external IDs, cache expiration, duplicate prevention, and migration from the legacy table shape.

## Related Issues [optional]

- Closes #3389

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [x] Yes
- [ ] No


badja-dev (commented):
`get_episode_by_id`'s cold-miss path calls `get_show` and `get_season` for every season of the show, and `get_season`/`get_show` can raise a bare `TMDbException` (not wrapped in Failed) in some paths. Previously a single episode lookup risked that; butnow a miss touches the whole show,widening the scope for an unhandled exception. I don't think this is new, but worth handling.

`Plex.get_tmdb_episode_id` duplicates the `urlparse(guid_tag.id)` GUID-parsing pattern already used in `get_watchlist` instead of factoring out a shared helper - cosmetic problem, can be ignored. Worth knowing about

antwanchild:
Addressed the substantive retry concern in ec63f613. The cold episode-map boundary now catches both `TMDbException` and exhausted Tenacity `RetryError` from show/season loading and converts them to `Failed`, allowing `Plex.get_tmdb_episode()` to use its positional fallback. Added separate show-retry and season-client-failure regression tests. Focused: 127 passed; full suite: 808 passed. I left the GUID parsing refactor scoped out as suggested.

badja-dev (approved):
LGTM

PR #3409, opened 2026-07-22
Standardize transient TMDb retry handling

antwanchild:
## What type of PR is this?

- [X] Bug Fix
- [ ] Feature/Tweak
- [ ] Refactor
- [ ] Documentation Update
- [ ] Build or CI Change
- [ ] Other

## Description

Standardizes handling of transient TMDb failures. The original trigger was a `(502 [Bad Gateway])` response with TMDb status code `43` during the `ITVX Shows` collection, which escaped as an unknown error and produced repeated tracebacks.

### Retry classification

| Condition | Behavior |
|:--|:--|
| Connection errors and timeouts | Retry up to six attempts |
| HTTP `408`, `429`, and `5xx` responses | Retry up to six attempts |
| TMDb service codes `9`, `24`, `25`, `43`, and `46` | Retry up to six attempts |
| HTTP `401`/`404`, Kometa `Failed`, and programming exceptions | Terminal; do not retry |
| Transient failure after all attempts | Raise `Unavailable`, categorized as a `ServiceError` |

Structured `response.status_code` and exception `status_code` attributes take precedence over regex parsing of exception messages.

### Affected request paths

| Path | Change |
|:--|:--|
| Existing TMDb request wrappers | Use the shared transient-failure policy |
| Discovery pagination | Construct the discovery object once and retry only `get_results`, preserving its page cache across attempts |
| Lazy movie, show, and episode hydration | Keep property-triggered API calls inside the retry boundary |

After retries are exhausted, the collection runner records a Service Error instead of reporting a generic unknown error with a traceback.

### Testing

| Validation | Result |
|:--|:--|
| Regression coverage | Added for structured-status precedence, transient and terminal failures, pagination reuse, discovery recovery/exhaustion, and lazy hydration |
| Pyright baseline | Passed: 5 baseline errors, 5 current errors, no regressions |
| Black | Passed |
| Focused pytest | Not run locally because the host lacks the pinned runtime dependencies and dependency installation was unavailable |

## Related Issues [optional]

None.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

### DIFF
--- a/.github/.wordlist.txt
+++ b/.github/.wordlist.txt
@@ -31,6 +31,7 @@ atmos
 atres
 avenir
 Babygirl
+backend
 backoff
 bafta
 bambara
@@ -454,6 +455,7 @@ TomatoRotten
 tooltip
 tooltips
 traceback
+tracebacks
 trakt
 trakt's
 transcoded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Treat TMDb connection failures, timeouts, HTTP 408/429/5xx, and backend-unavailable responses as transient service failures: retry discovery, pagination, item requests, and lazy object hydration consistently; emit concise retry warnings instead of tracebacks; and surface exhausted retries as a service-unavailable error.
 - `modules/plex.py`: standardize Plex API retry handling so known 400/404/401 responses and Kometa `Failed` exceptions remain terminal, while exhausted transient retries re-raise their underlying exception instead of becoming opaque `RetryError`; collection move failures also include the item title and original Plex response, and `query_collection`/`get_actor_id` convert terminal Plex responses into contextual Kometa errors.
 - Drop-in replacement for the jikan api
 
--- a/modules/tmdb.py
+++ b/modules/tmdb.py
@@ -142,6 +142,8 @@ def _is_transient_tmdb_exception(exception):
         exception_status = getattr(current, "status_code", None)
         structured_statuses = []
         for status in (response_status, exception_status):
+            if status is None:
+                continue
             try:
                 structured_statuses.append(int(status))
             except (TypeError, ValueError):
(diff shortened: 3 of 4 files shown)


YozoraXCII:
@badja-dev to review

badja-dev (changes requested):
In `_is_transient_tmdb_exception`, it does `http_status = int(http_match.group(1)) if http_match else response_status` - the fragile string-regex parse is tried before the structured `response.status_code`/`exception.status_code` attributes, when it should be the other way around - should always do structured before a regex imo

Wrapping the whole discover call in retry has a hidden cost - `self.TMDb.discover_movies/discover_tv_shows(**attrs)` is called inside the retried function, so a transient failure recreates the pagination object from scratch on retry, losing tmdbapis's own per-instance page cache (`_page_storage` in `TMDbPagination.load_page`).
I tested this with the same retry config and tmdbapis's actual caching logic: one failure on page 5 of an 8-page discover result caused 13 total page fetches instead of 8, re-fetching pages 1-4 for no reason.
The later in a large paginated result the failure lands, the worse this gets. Fix: call `discover_movies`/`discover_tv_shows` once outside the retry boundary and only wrap `results.get_results(amount)`, so `load_pag`e's existing cache lets a retry resume from the failed page instead of restarting from page 1

antwanchild:
Thanks @badja-dev — both review points are addressed.

- Structured `response.status_code` and exception `status_code` attributes now take precedence over regex message parsing.
- The discovery object is now constructed once, with only `get_results` inside the retry boundary, so the existing pagination page cache is preserved across attempts.
- Regression tests cover structured-status precedence and verify that discovery is constructed once while `get_results` is retried.
- The branch has been rebased onto the latest `nightly`, and the Pyright baseline check passes at 5 → 5 with no regressions.

The implementation updates are in `c850c5af` with the typing follow-up in `aab5b18c`. Please re-review when you have a chance.

antwanchild:
@YozoraXCII this should be good now...

YozoraXCII:
@badja-dev happy with the changes?

badja-dev (approved):
LGTM

PR #3412, opened 2026-07-23
Fix smart-label collection deletion

antwanchild:
## What type of PR is this?

- [X] Bug Fix
- [ ] Feature/Tweak
- [ ] Refactor
- [ ] Documentation Update
- [ ] Build or CI Change
- [ ] Other

## Description

Fixes smart-label collection deletion failing with:

```text
AttributeError: 'Plex' object has no attribute 'batch_edit_tags'
```

### What we discovered

The failure was reproduced twice in a nightly `2.4.5-build2` run while Kometa attempted to delete the unscheduled `Halloween Movies` and `Christmas Movies` smart-label collections. Both tracebacks followed the same path through `CollectionBuilder.delete()`:

1. Find every item carrying the smart collection's label.
2. Call `self.library.batch_edit_tags(...)` to remove that label.
3. Delete the smart collection.

The batching work originally added this caller, but the unreleased tag/genre/rating batching implementation was subsequently reverted because it was not ready. This smart-label deletion caller was left behind even though `Plex.batch_edit_tags` no longer exists.

| Effect | Result |
|:--|:--|
| Label cleanup | Crashes before any labels are removed |
| Collection deletion | Never reached, so the unscheduled collection remains |
| Run reporting | Collection is recorded as an Unknown Error with a traceback |
| Repetition | Occurs once for every smart-label collection that enters this deletion path |

### Why this fix

This restores the previously supported behavior: iterate over the matching Plex items and call the existing `edit_tags` API for each one. It does not reintroduce the removed batching implementation or duplicate its unfinished tag-diff behavior.

A regression test now exercises the complete smart-label deletion path and verifies that:

- the label search uses the collection name and builder level;
- every matching item has the smart label removed through `edit_tags`;
- the collection is deleted only after label cleanup completes; and
- the builder records the collection as deleted.

### Testing

| Validation | Result |
|:--|:--|
| Pyright baseline | Passed: 5 baseline errors, 5 current errors, no regressions |
| Black | Passed |
| Focused pytest | Not run locally because the host lacks the pinned `arrapi` dependency required to import `modules.builder` |

## Related Issues [optional]

None.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Treat TMDb connection failures, timeouts, HTTP 408/429/5xx, and backend-unavailable responses as transient service failures: retry discovery, pagination, item requests, and lazy object hydration consistently; emit concise retry warnings instead of tracebacks; and surface exhausted retries as a service-unavailable error.
 - Add the affected URL and HTTP status to exhausted empty TVDb response warnings; limit empty documents to three attempts with short exponential delays while retaining six attempts and the existing delay for other transient failures; and open a run-scoped circuit after two distinct URLs exhaust retries so a systemic TVDb outage does not generate thousands of requests and hours of retry delays.
 - Stop ntfy from sending an access-test notification whenever Kometa initializes the ntfy client.
+- Restore per-item label removal when deleting smart-label collections, preventing an `AttributeError` from the removed `batch_edit_tags` API.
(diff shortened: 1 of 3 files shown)


antwanchild:
@badja-dev Would you mind reviewing this fix when you have a chance?

YozoraXCII:
Is this supposed to be touching ntfy stuff? @antwanchild

antwanchild:
> Is this supposed to be touching ntfy stuff? @antwanchild 

Ummm nope. Carrot merged my ntfy pr, and I need to rebase this pr to remove those

@YozoraXCII this is now rebased

PR #3416, opened 2026-07-24
Stop ntfy startup access-test notifications

### LINKED ISSUE #3415: Native ntfy integration sends an access-test notification on every run
## Version Number

2.4.4

## Branch

master

## Describe the bug

When Kometa's native ntfy integration is configured, Kometa sends a low-priority `Kometa - Testing ntfy Access` notification whenever the ntfy client is initialized. In a normal scheduled setup this results in an access-test notification on every Kometa run, even when only an event such as `error` is routed to ntfy.

This appears to happen because `Ntfy.__init__()` calls `self._test_url()` unconditionally:

https://github.com/Kometa-Team/Kometa/blob/master/modules/ntfy.py

There does not appear to be a configuration option to disable the recurring access test independently of the actual ntfy webhook events.

The original ntfy implementation PR documents that an access-test message is sent at priority 1, but the current ntfy configuration documentation does not explain that this occurs during every normal initialization/run:

- https://github.com/Kometa-Team/Kometa/pull/2475
- https://kometa.wiki/en/latest/config/ntfy/

## Steps to reproduce

1. Configure the root `ntfy` mapping with a valid server, topic, and token.
2. Route only an event such as `error` to `ntfy`.
3. Run Kometa normally more than once, either manually or on a schedule.
4. Observe a new `Kometa - Testing ntfy Access` notification for each run even when no configured event occurred.

## Minimal configuration

```yaml
ntfy:
  url: https://example.invalid
  token: redacted
  topic: redacted

webhooks:
  error: ntfy
  version:
  run_start:
  run_end:
  changes:
  delete:
```

## Expected behavior

A normal Kometa run should send only the webhook events explicitly enabled by the user. The access test could instead be limited to explicit configuration validation/test mode, or controlled by an option such as `test_on_start: false`.

## Actual behavior

The access-test notification is sent whenever the ntfy client is initialized, creating recurring notification noise for frequently scheduled Kometa runs.

## Logs

No error is produced. The behavior is deterministic from the current `Ntfy.__init__()` implementation and was reproduced with Kometa 2.4.4 on `master`.

antwanchild:
## What type of PR is this?

- [X] Documentation Update
- [ ] Other

## Description

Stops the native ntfy integration from publishing a low-priority `Kometa - Testing ntfy Access` notification every time Kometa initializes.

### Root cause

`Ntfy.__init__()` called `_test_url()` unconditionally. That method attempted to validate write access by publishing a real notification, so every normal Kometa run generated a message even when the user had enabled only a specific webhook event such as `error`.

### New behavior

| Stage | Previous behavior | New behavior |
|:--|:--|:--|
| ntfy initialization | Publish an access-test notification | Load and validate the configured values without making a network request |
| Configured webhook event | Publish the event notification | Unchanged |
| Invalid network, token, or topic write access | Detected by publishing the startup test | Reported when the first configured ntfy webhook is sent |

The ntfy-specific startup test has been removed rather than replaced with a new setting. Other notification integrations do not expose a recurring startup-publish option, and retaining one for ntfy would add configuration solely to preserve the unwanted behavior.

Notifiarr validates its API key using a read-only user-settings endpoint. ntfy topics can have separate read and write permissions, so a subscription/read request would not prove that Kometa can publish. The ntfy status log therefore now reports successful configuration instead of claiming a connection was tested.

### Testing

Regression coverage verifies that:

- constructing the ntfy client performs no POST; and
- configured event notifications continue to use the existing request path.

| Validation | Result |
|:--|:--|
| Python compilation | Passed |
| Black | Passed |
| Focused pytest | Could not collect locally because the host lacks the pinned `arrapi` dependency imported by the shared small-modules test file |

## Related Issues [optional]

- Closes #3415

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Treat TMDb connection failures, timeouts, HTTP 408/429/5xx, and backend-unavailable responses as transient service failures: retry discovery, pagination, item requests, and lazy object hydration consistently; emit concise retry warnings instead of tracebacks; and surface exhausted retries as a service-unavailable error.
 - Add the affected URL and HTTP status to exhausted empty TVDb response warnings; limit empty documents to three attempts with short exponential delays while retaining six attempts and the existing delay for other transient failures; and open a run-scoped circuit after two distinct URLs exhaust retries so a systemic TVDb outage does not generate thousands of requests and hours of retry delays.
+- Stop ntfy from sending an access-test notification on every run; the startup publish test is now opt-in with `ntfy.test_on_start: true`.
 - `modules/plex.py`: standardize Plex API retry handling so known 400/404/401 responses and Kometa `Failed` exceptions remain terminal, while exhausted transient retries re-raise their underlying exception instead of becoming opaque `RetryError`; collection move failures also include the item title and original Plex response, and `query_collection`/`get_actor_id` convert terminal Plex responses into contextual Kometa errors.
 - Drop-in replacement for the jikan api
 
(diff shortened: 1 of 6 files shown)


chazlarson (approved):
LGTM

PR #3427, opened 2026-07-25
Resolve media paths using the server's path style, not Kometa's

nickwolf:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Plex reports library paths using the separator of the machine running the server, which need not match the machine running Kometa. A Kometa install on Linux (for example the Docker image) pointed at a Plex server on Windows gets paths like `P:\Movies\Title\file.mkv`.

`os.path.dirname` only understands the separator of the platform Kometa itself is running on, so on Linux it finds no `/` in that string and returns `""`:

```python
>>> import os
>>> os.path.dirname(r'P:\Movies\Title (2024)\Title (2024).mkv')   # on Linux
''
```

Both call sites treat that empty result as "no location found":

- `modules/builder.py` logs `Plex Error: No location found for <title>: [...]` and skips the item
- `modules/operations.py` has the same `os.path.dirname` call in the `radarr_add_all_existing` / `sonarr_add_all_existing` path

The user-visible effect is that `add_existing`, `upgrade_existing`, `monitor_existing`, `item_radarr_tag` and `item_sonarr_tag` silently do nothing for the entire library, while the log fills with one error per item. On my library that was 55,866 errors in a single run.

**This is not a regression.** The same `os.path.dirname` call is present in v2.1.0, v2.3.1 and v2.4.0. It only appears once Kometa and Plex are on platforms with different path styles, which is why it typically shows up right after someone moves an existing bare-metal Windows install into Docker — which is exactly what happened here. Collections, posters and overlays are unaffected (they go through the Plex API and never touch the path), so the failure is easy to miss.

### The change

Adds `util.media_dirname`, which picks the path flavour from the path itself instead of from the host platform, and uses it at both call sites:

```python
def media_dirname(path):
    path = str(path)
    if ntpath.splitdrive(path)[0] or path.startswith("\\\\") or ("\\" in path and "/" not in path):
        return ntpath.dirname(path)
    return os.path.dirname(path)
```

Windows drive-letter paths and UNC paths route to `ntpath`; everything else keeps the current `os.path` behaviour, so POSIX-to-POSIX and Windows-to-Windows setups are untouched.

### Testing

Verified on the affected setup (Kometa in Docker on WSL2, Plex on Windows, library on `P:\`):

```
old  os.path.dirname -> ''                        | falsy: True
new  media_dirname   -> 'P:\Movies\Title (2024)'  | falsy: False
```

- Added 5 tests in `tests/test_util.py` covering Windows drive-letter paths, UNC paths, POSIX paths, the truthiness that the callers depend on, and non-`str` input.
- Full suite passes: `914 passed`.
- `black --check`, `isort --check-only`, `flake8 --max-line-length=256 --extend-ignore=E203,W503,E501`, `bandit` and `pyspelling` all clean.

## Related Issues [optional]

- Related Issue #
- Closes #

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `modules/plex.py`: standardize Plex API retry handling so known 400/404/401 responses and Kometa `Failed` exceptions remain terminal, while exhausted transient retries re-raise their underlying exception instead of becoming opaque `RetryError`; collection move failures also include the item title and original Plex response, and `query_collection`/`get_actor_id` convert terminal Plex responses into contextual Kometa errors.
 - Added Drop-in replacement for the jikan api as it is being deprecated
 - Fix the startup requirements version check so each package is compared against its own pinned requirement. Previously `v1`/`v2` leaked between loop iterations and the package name lookup was case-sensitive, causing bogus messages such as `GitPython version: 3.1.55 does not match expected: 1.4.14`, where the expected value came from an unrelated package.
+- Resolve movie folder paths using the path style reported by the media server rather than the platform Kometa runs on. A Linux or Docker install pointed at a Plex server on Windows received paths such as `P:\Movies\Title\file.mkv`, which `os.path.dirname` reduced to an empty string, logging `Plex Error: No location found` for every movie and silently disabling `add_existing`, `upgrade_existing`, `monitor_existing`, `item_radarr_tag`, and `item_sonarr_tag`.
 
 ## [v2.4.5] - 2026-07-22
 
--- a/modules/builder.py
+++ b/modules/builder.py
@@ -4819,7 +4819,7 @@ class CollectionBuilder:
             ):
                 if item.locations:
                     if self.library.is_movie:
-                        path = os.path.dirname(str(item.locations[0]))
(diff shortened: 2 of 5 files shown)


YozoraXCII:
@chazlarson could you review this one please

badja-dev (changes requested):
Logic is fully sound and tested my side.

Please however, sort out `def media_dirname(path):` - the `media_dirname` docstring is a multi-paragraph block, more verbose than the one-liners elsewhere in util.py. Needs a trim please

nickwolf:
> Logic is fully sound and tested my side.
> 
> Please however, sort out `def media_dirname(path):` - the `media_dirname` docstring is a multi-paragraph block, more verbose than the one-liners elsewhere in util.py. Needs a trim please

I'll revise/trim today, thanks for the review!

nickwolf:
Trimmed in ef07af3 - the docstring is now a one-line summary plus a two-line note on the why, same shape as `is_locked` just below it. The longer explanation is still in the commit message and the CHANGELOG entry. No logic changes; `tests/test_util.py` still passes (25 passed).

badja-dev (approved):
LGTM

PR #3434, opened 2026-07-28
feat: improved language codes support

### LINKED ISSUE #3433: Language flags skipped with non-standard language tags
### Version Number

2.4.5

### What branch are you on?

master

### Describe the Bug

Some titles have language tags of media stream / subtitles in a non-standard format like "en-US" / "es_ES" / "es-419" (latino spanish), etc. Currently, Komat can't handle this and therefore language flags are skipped for these titles. The fix is simple, match up until the ISO country code and ignore the rest.

Will open a small PR for the fix, hoping to get reviewed / merged asap.

```yml

```

### Logs

_No response_

szab100:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description
Kometa can't handle language tags of audio streams / subtitles when it is either a 3-letter (ISO-639-2) code (like 'ita', 'chi', etc) or locale-specific (like 'en-US' / 'es-419' / 'es-ES' etc) language code, starting with the 2-letter ISO-639-1 language code, while Plex supports any variations of these and many media titles use this, which currently get no language flag overlays from Kometa.

image

This PR is aiming to support any supported variations and derive it to the standard 2-letter ISO 639-1 language codes.

## Related Issues [optional]

- Related Issue #3433 
- Closes #3433 

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


antwanchild:
@szab100 PRs are not allowed on master-  please change to nightly

antwanchild:
@szab100 needs to rebase to nightly

szab100:
@antwanchild Done, PTAL.

Tested with a locally built image - the previously skipped titles (like the one on the screenshot above) now have the proper language flags.

defaults/overlays/languages.yml
```diff
@@ -193,9 +193,9 @@ templates:
     ignore_blank_results: true
     run_definition:
       - <<use_<<key>>>>
-    plex_search:
-      all:
-        <<search_attribute>>: <<key>>
+    plex_all: true
```
YozoraXCII, line 196: Is this going to increase runtimes for everyone? It will have to load in every item in the library rather than a targeted `plex_search`

szab100: Great question. Basically yes, however, plex_all doesn't multiply per language. The item list and per-item metadata reloads are both cached in-memory for the run, so the whole file does one full-library reload pass, not 80.

That said, there is a small added cost: every item now gets fully reloaded once per run (to read stream data), whereas plex_search let Plex filter server-side with no reload needed (although it is done 80x instead of once now due to caching). I had to move off plex_search because Plex's language filters are exact-match only, so a single query per language misses 3 letter and locale specific variants like ita/chi/es-419/es-MX.

There is a potential alternative solution that avoids the full search / reload: call Plex's lightweight listFilterChoices to see which distinct language variants actually exist, then run a targeted plex_search per language variant. Stays server-side, no reload, at the cost of ~80 small list requests instead of one full reload pass (cached during first pass). Likely better for very large libraries when only a subset of languages are selected for overlay, while for average library sizes the current solution’s true added cost should be small thanks to the heavy memoization. Either way, happy to switch if that’s preferred, if it works, it’s obviously more efficient, but it’s a bit more complex change.

szab100:
Based on @YozoraXCII's comment, I ended up creating a v2 of this PR (#3440) using a much more efficient approach: 
1. fetch filter values for audio & subtitle languages first (cached for subsequent configured languages)
2. group all supported language variants by the existing 2-letter lang codes
3. only perform search once per language (incl. variants using OR) that actually exist in the library

PR #3206, opened 2026-06-21
fix: episode audio/subtitle language plex_search returns fewer results than expected on show libraries

badja-dev:
## What type of PR is this?

- [X] Bug Fix (non-breaking change which fixes an issue)
- [ ] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [ ] Other

## Description

**Root cause:** `get_search_choices` in `modules/plex.py` builds a `choices` dict by looping over Plex filter entries and writing `choices[choice.title] = choice.key`. For show libraries, the `episode.audioLanguage` and `episode.subtitleLanguage` filter endpoints return multiple entries with the same display title but different keys - for example both `("English", "en")` and `("English", "en-US")`. Because the loop overwrites on each pass, whichever entry comes last wins. In practice the regional code (`en-US`, `es-ES`, `hi-IN`, etc.) lands last, so `choices["English"]` ends up as `en-US` instead of `en`. The search URL then becomes `episode.audioLanguage=en-US`, which only matches content explicitly tagged with that locale - not content tagged with the base code `en`. Most content is tagged with base codes, so the collection comes back almost empty.

**Fix:** In `get_search_choices`, when the filter is an episode language field and `use_title` is `False` (plex_search mode), the value stored under the title key is normalised by stripping the region suffix before writing - so `"English"` always resolves to `en` regardless of what order Plex returns the entries. The full regional code is preserved for direct key lookups, so `audio_language: en-US` still targets that specific locale.

**Real data - choices dict before and after (live Plex, episode.audioLanguage):**

```
Plex returns:  ('English', 'en')  then  ('English', 'en-US')

LOOKUP KEY     STOCK result    FIXED result
English        en-US           en             <- FIXED
english        en-US           en             <- FIXED
en             en              en             (same)
en-US          en-US           en-US          (same - direct lookup preserved)
```

**Real data - Kometa run against a 50-show TV library:**

```
OVERLAY                 BUG URL              FIX URL              BUG items  FIX items
audio_language English  audioLanguage=en-US  audioLanguage=en     8          50
audio_language Spanish  audioLanguage=es     audioLanguage=es     2          2  (already correct)
subtitle_language Span  subtitleLanguage=es-ES subtitleLanguage=es 14         12
```

[...]

PR #3440, opened 2026-07-29
feat: improved language codes support [v2]

### LINKED ISSUE #3433: Language flags skipped with non-standard language tags

szab100:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description
Kometa can't handle language tags of audio streams / subtitles when it is either a 3-letter (ISO-639-2) code (like 'ita', 'chi', etc) or locale-specific (like 'en-US' / 'es-419' / 'es-ES' etc) language code, starting with the 2-letter ISO-639-1 language code, while Plex supports any variations of these and many media titles use this, which currently get no language flag overlays from Kometa.

image

This PR is aiming to support any supported variations and derive it to the standard 2-letter ISO 639-1 language codes.

## Related Issues [optional]

- Related Issue #3433 
- Closes #3433 

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


szab100:
This is PR is using an efficient approach to solve the problem:

1. fetch filter values for available audio & subtitle languages first (cached & reused by all configured languages)
2. group all supported language variants by the existing 2-letter lang codes
3. only perform search once per language (incl. variants using OR) that actually exist in the library

defaults/overlays/languages.yml
badja-dev, line 1: all -> any is required now that a single configured language expands to multiple plex_search variants (an item only needs to match one).

modules/builder.py
badja-dev, line 1: The `if plex_search and attribute in (...)` branch is an `if/elif` - when it's true, `get_search_choices()`'s exact-key lookup is never reached, even as a fallback. A user filtering on an exact locale value already present in the library (e.g. `audio_language: es-419`) won't match, since `get_language_search_values` only indexes by base code. Suggest falling back to `search_choices` when `variants` is empty, or better, folding this into `get_search_choices` itself (see plex.py comment)

modules/plex.py
badja-dev, line 1: This overlaps with #3206 (episode-level region stripping, already in nightly/2.4.2) but reimplements normalisation as a separate cache/lookup rather than extending it. Recommend generalising the existing `is_episode_lang` stripping in `get_search_choices` to use `base_language_code()` instead of `split("-")[0]`, and dropping the show-only restriction - that covers movies too, keeps the exact-key fallback intact, and avoids two parallel definitions of "base language code" living in the codebase

tests/test_builder.py
badja-dev, line 1: Good coverage of the expansion path. Missing a case for a user passing an exact locale/3-letter value directly (e.g. `es-419`) that already matches a library choice verbatim - see the bypass issue noted on `builder.py`

tests/test_plex.py
badja-dev, line 1: Solid unit coverage on `base_language_code`/`get_language_search_values` in isolation. Would also help to assert against the full `get_search_choices` bypass scenario mentioned above

badja-dev (changes requested):
Noted in #3433 - this was already semi-resolved in #3206, but does need further expanding as #3206 was pushed in `master` 2.4.2.

It's a real improvement over what we have (movies + shows, proper 3-letter/locale handling via `langcodes` instead of string-splitting). Two things before merge:

1. This overlaps with #3206, which already fixed the show-library case in nightly/2.4.2 by stripping region from episode-level language choices in `get_search_choices`. As written, this PR doesn't extend that fix - it adds a parallel path in `validate_attribute` that fully bypasses `get_search_choices` for `audio_language`/`subtitle_language` + `plex_search`. Suggest consolidating: generalise the existing `is_episode_lang` logic with `base_language_code()` rather than introducing a second mechanism. Cleaner, and avoids leaving dead code behind
2. As a side effect of bypassing `get_search_choices`, a user who configures an exact locale value that's already a valid library choice (e.g. `audio_language: es-419`) won't resolve, since the new cache only indexes by base code with no fallback to an exact match. Worth a test case and a fix (or fixed naturally if consolidated per #1)

defaults/overlays/languages.yml
szab100: Good catch, and already fixed in the recently pushed commit: plex_search now uses `any:` for the language template. Confirmed with a test that multiple expanded variants OR together correctly (an item matches if any variant matches).

modules/builder.py
szab100: True, thanks for flagging. Fixed it a level down rather than adding a `search_choices` fallback here: `get_language_search_values()` now builds an exact-key index alongside the base-code groups, and checks it first whenever the requested code differs from its own base form (e.g. es-419 <> es). So `audio_language: es-419` resolves to exactly that variant, while a bare `audio_language: es` still expands to every variant sharing that base, including cases where es is also itself a literal choice. Kept the check inside `get_language_search_values` rather than falling back to `get_search_choices` so we don't reintroduce the uncached `listFilterChoices` call for this path. Added regression tests for both the exact-match and the "base code wins even if also literal" cases.

tests/test_builder.py
szab100: Sure, added `test_exact_locale_value_passes_through_as_a_single_variant` in `TestValidateAttributeLanguage` covering this end-to-end through `validate_attribute`.

tests/test_plex.py
szab100: Added directly at the `get_language_search_values` level rather than re-testing the bypass itself. Also added `TestGetSearchChoicesEpisodeLanguage` covering the `base_language_code()` swap in `get_search_choices` (locale tag stripping, 3-letter code handling that the old split missed, and confirming movie libraries are unaffected by that branch).

modules/plex.py
szab100: Agreed on unifying the normalization. `get_search_choices's` episode-language branch now calls `base_language_code()` instead of `split("-")[0]`, so it also catches 3-letter ISO 639-2 codes the old split missed (e.g. `spa` vs. `es`), so same intent as #3206 but unified and broader support.

I traced through where this branch actually still executes before deciding on the show-only (`is_episode_lang`) scope, and it's quite narrow. The branch only fires when `not use_title and is_episode_lang`. Since the new `plex_search` path for `audio_language`/`subtitle_language` now bypasses `get_search_choices` entirely (goes through `get_language_search_values` instead), and `filters:` on shows forces `use_title=True` (blocking the branch regardless), the only path left where this code executes at all is `audio_language.regex`/`subtitle_language.regex` combined with `plex_search: true`, on show libraries only. Dropping the show-only restriction would just extend that same narrow combination to movie libraries too, not something tied to this issue, and not something `filters:` users would ever hit either way. Given how thin the benefit is, I've left the show-only scope as-is rather than expanding it for symmetry alone. Happy to revisit if you know of a concrete case this misses.

badja-dev: I'll test the regex in the morning - everything else LGTM

YozoraXCII:
@badja-dev happy?

badja-dev (changes requested):
Traced the actual consumers rather than re-reading the code, and added two tests to confirm: `.regex` (builder.py's tag_attributes branch) discards `get_search_choices`'s `choices` dict and only reads `names`, which is built from the raw `choice.title`/`choice.key`, never `title_value`. So even in the "surviving caller" scenario, the stripped value is computed but never read

The other caller (`filters:`) always passes `title=True`, which blocks the branch too

Net result: nothing in the codebase consumes the `is_episode_lang` stripping anymore, regardless of scope. Tests: `choices["Spanish"] == "es"` while `names` still has raw `es-ES`/`es-MX`; and an end-to-end `.regex` + `plex_search=True` call returns the raw key, not the stripped one

Suggest just removing the `is_episode_lang`/`title_value` branch in this PR rather than debating scope, since it has no observable effect either way

szab100:
> Traced the actual consumers rather than re-reading the code, and added two tests to confirm: `.regex` (builder.py's tag_attributes branch) discards `get_search_choices`'s `choices` dict and only reads `names`, which is built from the raw `choice.title`/`choice.key`, never `title_value`. So even in the "surviving caller" scenario, the stripped value is computed but never read
> 
> The other caller (`filters:`) always passes `title=True`, which blocks the branch too
> 
> Net result: nothing in the codebase consumes the `is_episode_lang` stripping anymore, regardless of scope. Tests: `choices["Spanish"] == "es"` while `names` still has raw `es-ES`/`es-MX`; and an end-to-end `.regex` + `plex_search=True` call returns the raw key, not the stripped one
> 
> Suggest just removing the `is_episode_lang`/`title_value` branch in this PR rather than debating scope, since it has no observable effect either way

Got it, sure, thanks a lot for looking into this.

badja-dev (approved):
LGTM

PR #3383, opened 2026-07-18
perf: batch Plex writes and add in-run caching to reduce redundant API/network calls (part 2)

badja-dev:
## What type of PR is this?

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [X] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [ ] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

Reduces the number of Plex API calls and repeated network/lookup work made during a run, without changing behavior.

Plex writes that were previously one API call per item are now batched: label/genre sync, `item_critic`/`audience`/`user_rating` updates, and title-parentheses removal now merge into shared `saveMultiEdits()`/`batchMultiEdits()` calls (respecting `plex_bulk_edit_batch_size`) instead of one `edit_tags()`/`editField()`/`editTitle()` call per item. Label/genre and rating writes for the same item are further merged into a single PUT instead of one per attribute type. The overlay `Overlay` label add is batched the same way.

In-run caching/memoization was added for: `cache.py`'s ID-map queries (`query_tmdb_to_tvdb_map` and siblings, `query_guid_map`/`update_guid_map`), `check_filter`'s plain item-attribute reads and media-derived number filters (channels/height/width/aspect/versions/audio_language/subtitle_language/duration), and per-item seasons/episodes/albums/tracks sub-item listings via `cached_item_subitems`.

Reduced network/image work: `Requests.get_image()` is memoized by URL for the run, validate-only image URL checks use HEAD instead of GET, validate-only checks are skipped entirely for Kometa's own GitHub-hosted assets, and fresh local overlay images are reused instead of being re-fetched from GitHub on every call. Read-only YAML loads now use ruamel's safe loader.

Smaller wins: log file writes are now buffered instead of flushed after every line (`WARNING` and above still flush immediately), forced Plex reloads for tag filters checked back-to-back in `check_filters` are deduped, the overlay backup folder is snapshotted once per run instead of checked per item, and `check_for_var`'s variable-resolution loop early-exits once a pass makes no substitutions.

Also includes two small correctness fixes found while making these changes: `get_image()` now consistently routes through `self.get()`/`self.head()` so it keeps their `@retry` backoff behavior, and Plex batch writes now group items by Plex type first so `batchMultiEdits()` never receives a mixed show/season batch.

[...]

PR #3451, opened 2026-07-31
feat: concurrency/threading support - shared thread pool, cache RLock, prefetch defaults (part 3)

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Adds a shared `ThreadPoolExecutor` and a `settings.threading` config block (`workers`,
`prefetch_collection_children`, plus the experimental `tmdb_pages`/`parallel_sources`), then uses it for
the two changes that actually paid for themselves in measurement, plus three unrelated correctness bugs
found along the way via this project's own request-census tooling.

Builds on #3383 (batched Plex writes, in-run
caching), which has already merged into `nightly`. This PR is rebased directly onto current `nightly` -
the diff here is just this branch's own commits.

**What's on by default now:**
- `workers: 4` (was effectively unthreaded) and `prefetch_collection_children: true` - the cache is now
  RLock-guarded (`_LockedConnection`) so concurrent worker threads can share one SQLite connection
  safely; a collection's `sync_collection` "what to remove" lookup and PMS's comma-separated
  `/library/metadata/{ids}` batch-read are both deferred to the thread pool instead of blocking the main
  loop.
- Three independent correctness fixes: a bare `self.obj` truthy check that silently triggered a full
  Plex `items()` fetch through `Collection`/`Playlist.__len__` on any falsy-looking-but-real collection;
  a `childCount` `None` follow-on that surfaced once the above stopped masking it; and
  `get_item_display_title()` unconditionally fetching a parent's `titleSort` even when the caller didn't
  ask for sorted output.

**What's built but off by default** (measured, not stripped, in case the calculus changes on a bigger
library): concurrent `gather_ids` with per-service locks (`parallel_sources`, structurally ~0% gain -
none of the measured baseline's builder groups actually mix services within one collection). A separate
cross-collection pipelining experiment (`pipeline_depth`) was also tried and measured 1.8-4% *slower*
from GIL contention between background response parsing and main-thread Plex work - that one was
dropped from this branch entirely rather than kept behind a toggle, since it added real rebase/maintenance
surface for a change that never won.

**Why `workers: 4`/`prefetch_collection_children: true` as defaults, not just opt-in:** both only touch
work that was already safe to parallelize - a read-only Plex lookup and a shared cache connection now
guarded by an RLock - not the collection-building logic itself, so the risk profile is "more concurrent
I/O," not "different results." Zero Plex/network call-count regression across every A/B in this branch's
history (churn-loop and victory-lap alike), and the full test suite (1339 tests) passes unchanged. The
one real tradeoff is more simultaneous outbound requests, called out below.

## Gains measured

Full victory-lap run (5-pair A/B, standard test libraries, 10%-churn) put this branch's tip **41.3%
faster than stock nightly** (347.9s vs 592.5s mean) - the cumulative effect of everything above measured
together against a real feature-complete config, not summed from individual bake-offs. That number is
from the pre-rebase branch tip; this branch has since been rebased onto current `nightly` to pick up
everything merged since then (including this PR's own base, #3383), and hasn't had a fresh A/B pass
post-rebase yet. Treat 41.3% as the demonstrated order of magnitude, not a number to cite verbatim for
current `nightly` - happy to re-run the victory-lap comparison during review if that's useful.

## What users need to know

- Zero-config behavior changes: a config with no `settings.threading` block now runs with `workers: 4`
  and `prefetch_collection_children: true` instead of the old inert defaults. Anyone who already sets
  these explicitly is unaffected.
- More concurrent worker threads means more simultaneous outbound requests (Plex, TMDb, Trakt, etc.) at
  once - worth knowing if a library's Plex server or an external service is rate-limit-sensitive.
  `workers` is user-configurable if that matters for a given setup.
- `parallel_sources`/`tmdb_pages` stay off and undocumented for now - bake-off-measured as not worth it
  at steady state, kept behind their toggles rather than deleted.

## Testing

Full `pytest` suite: 1339 passed. `black`/`isort`/`flake8` clean. `pyright` baseline check passes with a
net improvement (8 -> 5 errors; two unrelated pre-existing gaps fixed along the way - see commit history
for detail). Churn-loop A/B methodology and full commit-by-commit history detail:
`threading-experiment-plan.md` in the "Kometa issues and Dev" project (not part of this repo).

## Related Issues [optional]

- Builds on [[perf: batch Plex writes and add in-run caching to reduce redundant API/network calls (part 2) #3383](https://github.com/Kometa-Team/Kometa/pull/3383)](https://github.com/Kometa-Team/Kometa/pull/3383) (merged), which itself built on [perf: add `--timings`/`KOMETA_TIMINGS` Runtime Flag for Timing Instrumentation (part 1) #3382](https://github.com/Kometa-Team/Kometa/pull/3382) (merged).
- Three of this branch's commits (`self.obj`, `childCount`, `titleSort` fixes) are independent
  correctness bugs, not threading-specific - candidates for their own small PRs if reviewers would
  rather land those separately from the threading work.

- [x] Yes
- [ ] No
- [ ] Not Applicable

Added a `threading` attribute block to `docs/config/settings.md` (documenting `workers` and
`prefetch_collection_children`; the experimental sub-attributes are intentionally left undocumented in
prose, matching their off-by-default/unsupported status) and a commented-out example in
`config/config.yml.template`.

- [x] Yes
- [ ] No
- [ ] Not Applicable

`settings.threading` added to `json-schema/config-schema.json`, including the undocumented experimental
sub-attributes (typed but undescribed) so users who do set them still get valid schema validation instead
of an `additionalProperties` error.

- [x] Yes
- [ ] No

### DIFF
--- a/.pyright-baseline.json
+++ b/.pyright-baseline.json
@@ -1,12 +1,12 @@
 {
   "_comment": "Ratcheting pyright baseline. DO NOT hand-edit unless you know what you're doing. Regenerate with: python scripts/pyright_baseline.py --regenerate",
-  "total_errors": 8,
+  "total_errors": 7,
   "per_file_errors": {
     "kometa.py": 1,
     "modules/builder.py": 1,
     "modules/config.py": 1,
     "modules/meta.py": 1,
-    "modules/operations.py": 2,
+    "modules/operations.py": 1,
     "modules/plex.py": 2
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support the `folder_location` Plex search option in music-library track builders.
 - Add a `settings.threading` config block (`workers`, `tmdb_pages`, `parallel_sources`, `prefetch_collection_children`) backed by a shared `ThreadPoolExecutor`, used to defer non-Plex `gather_ids` work and Plex item-reload batching off the main collection loop.
-- Add a `--dcl`/`--delete-collections-labels` CLI flag (double-dash only - `-dcl` collides with `-d`/`--divider`) that deletes all collections, same as `-dc`, and for any deleted collection whose title matches an existing Plex label of the same name (the Smart Label default), batch-removes that one label from just the items that have it - instead of `-dl`'s full library-wide, one-item-at-a-time label wipe.
+- Add a `--delete-collections-labels` CLI flag (no short form - would collide with `-d`/`--divider`) that deletes all collections, same as `-dc`, and for any deleted collection whose title matches an existing Plex label of the same name (the Smart Label default), batch-removes that one label from just the items that have it - instead of `-dl`'s full library-wide, one-item-at-a-time label wipe.
 
 ### Fixed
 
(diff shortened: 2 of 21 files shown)


YozoraXCII:
@badja-dev worth resolving these conflicts to get this merged?

Issue #3245, opened 2026-06-24
Dependabot triage: 15 of 16 open alerts are scanner lag, 1 is a Low-impact docs-only CVE


chazlarson:
**Date:** 2026-06-23
**Branch scanned:** `nightly` (default)
**Source:** GitHub Dependabot alerts API + `uv.lock` reality-check

---

## TL;DR

| Headline | Reality |
|---|---|
| Push-time warning says "16 vulnerabilities (10 high, 6 moderate)" | Technically true |
| Actually actionable today | **1 medium-severity alert** |
| Already fixed in `uv.lock` (v2.4.2 release) | **15 of 16 alerts** |
| Why Dependabot still shows them | It hasn't re-scanned since the lockfile bump |

**The takeaway:** the security tab looks alarming but it is, almost entirely, scanner lag. The one real residual is a *Low impact* CVE in a docs-build-only dependency that we control all input to.

---

## What I'm asking the maintainers

1. **Confirm the threat-model read on the one real CVE** (mkdocs-include-markdown-plugin, "Low impact", docs-build-only). If you concur, my preference is **Option C: document and dismiss the alert** with a short justification in the security tab.
2. **If you'd rather close the gap entirely**, I'm happy to PR a fork bump (Option B is smallest: cherry-pick just commit `7466d67` from upstream into `Kometa-Team/mkdocs-include-markdown-plugin@master`). Requires someone with write access to that repo to either land it or grant access.
3. **No action needed on the other 15 alerts** — they'll auto-close on the next Dependabot re-scan triggered by any future `uv.lock` change, or on the scheduler's daily cadence.

This issue is mostly informational — closing once the one CVE is decided + Dependabot re-scans the other 15.

---

## Background

After pushing to `nightly` this week, every `git push` emits:

```
remote: GitHub found 16 vulnerabilities on Kometa-Team/Kometa's default branch
remote: (10 high, 6 moderate). To find out more, visit:
remote:      https://github.com/Kometa-Team/Kometa/security/dependabot
```

That's enough noise to want a real triage. The alerts data is scoped to `uv.lock`, so the question is whether the lockfile-pinned versions are actually below the patched threshold.

---

## Per-package reality check

For each open alert, we compared what's pinned in `uv.lock` to the `first_patched_version` reported by the advisory:

```
PACKAGE                          IN LOCK    VULN RANGE                PATCHED    STATUS
[...]

PR #3226, opened 2026-06-22
2.4.2

YozoraXCII:
## What type of PR is this?

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ ] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [ ] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

_Please replace this line with a meaningful description of your PR. What does it do? Why? Has it been tested? What were the results?_

## Related Issues [optional]

- Related Issue #
- Closes #

## Have you updated the Documentation to reflect changes (if necessary)?

- [ ] Yes
- [ ] No
- [ ] Not Applicable

## Have you updated the JSON Schema files (if necessary)?

- [ ] Yes
- [ ] No
- [ ] Not Applicable

## Have you updated the CHANGELOG.md?

- [ ] Yes
- [ ] No

PR #3475, opened 2026-08-04
chore: reintroduce uv.lock for dependency graph and Dependabot tracking

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [x] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

Reintroduces uv.lock, which was deleted and gitignored in #3226 as part of a partial switch to uv for local dev environments. CI never adopted uv (it still installs via pip install -r requirements.txt), so removing the lockfile did not affect builds, but it did silently break GitHub's dependency graph for this repo. Without a lockfile, Dependabot has no way to determine the resolved commit for git-based dependencies, which is why the mkdocs-include-markdown-plugin security alert (GHSA-v39m-5m9j-m9w9, Kometa-Team/Kometa#3245) has stayed open since June even after the underlying fix merged upstream in the plugin itself. It also explains why the two open Dependabot version-bump PRs tagged python:uv (#3222, #3223) have been stuck, since they were trying to update a file that no longer exists.

This PR regenerates uv.lock from the current pyproject.toml with uv lock, verified with uv sync --group dev (98 packages, installs cleanly). The plugin dependency now pins to its current commit, which already contains the CVE fix.

Ongoing cost worth flagging: mkdocs-include-markdown-plugin tracks the git branch master rather than a fixed version, so this lockfile's pin for it will drift stale again whenever that branch moves. Worth deciding whether to add a scheduled job to regenerate uv.lock periodically, or accept manual regeneration when needed.

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [ ] Yes
- [x] No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,9 +109,6 @@ ipython_config.py
 #   install all needed dependencies.
 #Pipfile.lock
 
-#uv
-uv.lock
-
 # PEP 582; used by e.g. github.com/David-OConnor/pyflow
 __pypackages__/
 
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,1953 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
+
+[[package]]
+name = "anyio"
+version = "4.14.2"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "idna" },
+    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
+]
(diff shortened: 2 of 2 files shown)


YozoraXCII:
Not sure what we're doing with this following internal discussion?

badja-dev:
> Not sure what we're doing with this following internal discussion?

@YozoraXCII This directly connects to #3245. When uv.lock last existed, it triggered that "16 vulnerabilities" warning, but Chaz found 15 were already fixed, Dependabot just hadn't rescanned yet. The only real one was mkdocs-include-markdown-plugin, and even that's low risk in practice (docs build only, content we write ourselves). His suggestion was to just dismiss that one alert in the Security tab with a note, rather than keep a lockfile that re-triggers this same noise every time it changes

Also, the actual fix for that CVE (CVE-2025-59940 / GHSA-v39m-5m9j-m9w9) is already merged into our plugin fork via #1, so it's patched either way

I think it makes more sense to close this PR and dismiss the one alert manually instead of reintroducing uv.lock. Open to pushback if people want full lockfile tracking, but this PR isn't the way to get there without causing the same alert storm again

PR #3476, opened 2026-08-04
Fix overlay reapplication after mass poster updates

YozoraXCII:
## Description

Fixes an issue where `mass_poster_update` could reset an overlaid poster, remove the `Overlay` label, and then cause the following overlay phase to incorrectly skip reapplying overlays.

After removing the label, Kometa now invalidates the affected item’s in-run Plex cache. The overlay phase then reloads the current Plex item state, detects the missing label, and reapplies the configured overlays in the same run.

## Changes

- Invalidate the affected in-memory item and filter-attribute cache after a poster reset removes `Overlay`.
- Add regression coverage for the cache invalidation behavior.
- Add an Unreleased changelog entry.

## Testing

- `uv run pytest tests/test_plex.py`
  - 126 passed

## Related issue

Fixes the alternating overlay/reset behavior when `mass_poster_update` runs before overlays.

### DIFF
--- a/modules/plex.py
+++ b/modules/plex.py
@@ -2077,8 +2077,6 @@ class Plex(Library):
                     logger.info(f"{text} | Reset from {location}")
                 if poster and "Overlay" in [la.tag for la in self.item_labels(item)]:
                     logger.info(self.edit_tags("label", item, remove_tags="Overlay", do_print=False))
-                    # The Overlay label was just changed on Plex. Do not let the overlay phase reuse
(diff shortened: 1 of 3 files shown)


YozoraXCII:
@badja-dev could you check over this one

modules/plex.py
badja-dev, line 1: I like, no complaints here except 2 line comment for a 3 line fix :D

badja-dev (approved):

PR #3496, opened 2026-08-12
Add MDBList collection synchronization

YozoraXCII:
## What type of PR is this?

- [] Documentation Update
- [ ] Other

## Description

Adds `sync_to_mdb_list` support for synchronizing Kometa collections with MDBList static lists.

The feature:

- Finds lists by exact name through the MDBList API.
- Creates the list automatically when it does not exist.
- Warns and stops when multiple lists have the same exact name.
- Supports `sync` mode (default), which adds and removes items to match the collection.
- Supports `append` mode, which only adds items.

Examples:

```yaml
collections:
  Recently Added:
    plex_search:
      all:
        added: 30
    sync_to_mdb_list: Recently Added
```

```yaml
collections:
  Favourite Shows:
    plex_search:
      all:
        user_rating.gte: 8.0
    sync_to_mdb_list:
      name: Favourite Shows
      mode: append
```

## Related Issues [optional]

- Related Issue #
- Closes #

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [] No

---
<a href="https://app.blacksmith.sh/Kometa-Team/codesmith/Kometa/pr/3496"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789127154&installation_model_id=431096&pr_number=3496&repository=Kometa-Team%2FKometa&return_to=https%3A%2F%2Fgithub.com%2FKometa-Team%2FKometa%2Fpull%2F3496&signature=9fc536d35dc0dd9cdf52e45d5897d266c6f362b7e2b5ddf7aa150e3907c74622"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/files/settings.md
+++ b/docs/files/settings.md
@@ -115,7 +115,7 @@ collections:
 If a list with the name specified exists, it will be used. If no list exists, one will be created.
 The default `mode` is `sync`, which adds missing items and removes items no
 longer present in the collection. Use `append` to add items without removing
-existing list entries. Movies, shows, seasons, and episodes are supported.
+existing list entries. Movies and shows are supported.
 
 ## Smart Label Definitions
 
--- a/docs/files/settings.md
+++ b/docs/files/settings.md
@@ -35,6 +35,7 @@ tags:
   - tmdb_birthday
   - changes_webhooks
   - sync_to_trakt_list
+  - sync_to_mdb_list
   - sync_missing_to_trakt_list
   - run_definition
   - default_percent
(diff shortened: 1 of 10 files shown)


docs/files/settings.md
badja-dev, line 1: New "MDBList Sync Example" section states "Movies, shows, seasons, and episodes are supported." I believe this is wrong... `sync_mdb_list()` in `builder.py` only handles `Movie` and `Show`. Either fix the docs to match current scope or implement season/episode support before merging

modules/builder.py
badja-dev, line 1: - Validation of `sync_to_mdb_list` (string or dict with `name`/`mode`) mirrors `sync_to_trakt_list` and looks good. Mode is restricted to `sync`/`append` with a clear `BuilderValidationError` on bad input, LGTM
- `sync_mdb_list()` only has branches for `Movie` and `Show`. There is no `Season` or `Episode` handling, unlike `sync_trakt_list()` which supports all four. This conflicts with the docs added in this same PR (see `docs/files/settings.md`) which claim season and episode support
- For shows, the code calls `self.config.Convert.tvdb_to_tmdb(pl_library.show_rating_key_map[item.ratingKey], fail=True)`. With `fail=True`, an unmapped show raises `MappingConvertError`, which is not caught locally. It propagates to the outer per-collection handler in `kometa.py`, which means one unmapped show can abort the rest of that collection's item-detail processing (custom sort, notifications, anything else queued in that block), not just the MDBList sync. `sync_trakt_list()` avoids this failure mode entirely since Trakt accepts tvdb ids natively and needs no conversion. Recommend wrapping the per-item conversion in a try/except that logs and skips the item instead of failing the whole block
- No `sync_missing_to_mdb_list` equivalent was added. May be intentional if MDBList has no similar scope

modules/mdblist.py
badja-dev, line 1: - `_request()` now supports POST via a new `json_data` param, encoding `apikey` onto the URL as a query string and sending the JSON body separately, LGTM
- Ambiguous list name handling only logs a warning and returns silently rather than raising `Failed`. Compare to how trakt is validated elsewhere with a hard error. A bad collection here will quietly no-op every run with nothing surfaced beyond a log line
- In sync mode, existing list contents are fetched via `get_tmdb_ids("mdblist_list", {"url": f"{base_url}{owner}/{list_slug}"})`, where `owner` and `slug` are pulled from the `lists/user` response by trying several possible key names (`username`/`user`/`owner`, `slug`/`list_slug`). If the real API response doesn't use any of those keys, `owner`/`slug` come back `None` and removal is silently skipped, so sync mode quietly degrades to append mode with no warning to the user
- Response parsing in `update_items()` has the same guessing pattern: `counts = result.get(action + "ed", result.get("removed", {}))`, plus fallback between `not_found`/`notfound`. Reads like it was written against an assumed response rather than a real one
- `int(str(item_id).split("_")[0])` in the payload builder is a bit guff. The caller in `builder.py` only ever produces plain movie/show tmdb ids, never composite ids with an underscore, so this split currently does nothing and just obscures what input format is actually expected... I think

kometa.py
badja-dev, line 1: LGTM

badja-dev (changes requested):
Good work, requires a bit of documentation cleanup and logical operation changes - will review again once complete

modules/builder.py
YozoraXCII: The documentation now accurately limits MDBList sync support to movies and shows.

A TVDb→TMDb conversion failure for an individual show is now caught, logged as a warning, and that item is skipped so the rest of the collection processing continues.

Will look into the missing in a follow-up PR potentially

modules/mdblist.py
YozoraXCII: Duplicate exact-name matches now raise a Failed error rather than silently returning.

Sync mode now fails clearly if the API response does not provide enough metadata to determine the list URL, rather than quietly behaving like append mode.

Removed the unnecessary underscore splitting from TMDb IDs.

Removed the speculative response-count parsing; the raw API response is trace-logged while progress is tracked from the submitted payload.

modules/builder.py
badja-dev: matching guard in `load_collection_items()` (not `self.items` and not `self.sync_to_mdb_list`) has the same over-broad scope, same fix needed

kometa.py
badja-dev: `arr_mdb_list_sync` guard change (and not `builder.sync_to_mdb_list`) applies to all sync collections, not just Arr ones - bad Plex filter now silently wipes the MDBList list in sync mode

badja-dev (changes requested):
PR is close but not mergeable yet - three of four round-one blockers are cleanly fixed (doc overclaim, whole-collection abort on unmapped shows, silent ambiguous-name no-op), but the new Arr-to-MDBList commit introduces a genuine data-loss risk (item 9: an over-broad "no items found" guard change in both kometa.py and builder.py lets any misconfigured `sync_to_mdb_list` collection silently wipe the external list instead of erroring) and ships with zero tests for that new path. Fix the guard scope, add a test proving it, then good to go

kometa.py
YozoraXCII: The empty-result bypass is now limited to actual Arr tag-list MDBList syncs, identified by mdb_list_arr_ids is not None. Standard sync_to_mdb_list collections now preserve the normal empty Plex-result failure path, preventing a bad Plex filter from silently clearing the MDBList list in sync mode.

I applied the same Arr-specific guard in load_collection_items() and added regression tests for both cases. Pyright found no diagnostics on the changed guards, and Black, isort, flake8, and the focused test suite pass.

modules/builder.py
YozoraXCII: The empty-result bypass is now limited to actual Arr tag-list MDBList syncs, identified by mdb_list_arr_ids is not None. Standard sync_to_mdb_list collections now preserve the normal empty Plex-result failure path, preventing a bad Plex filter from silently clearing the MDBList list in sync mode.

I applied the same Arr-specific guard in load_collection_items() and added regression tests for both cases. Pyright found no diagnostics on the changed guards, and Black, isort, flake8, and the focused test suite pass.

badja-dev (approved):
LGTM

PR #3509, opened 2026-08-16
Reduce MDBList API usage with bulk prefetching

chazlarson:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

MDBList charges one quota unit per request regardless of the number of IDs in that request. Kometa previously fetched MDBList metadata one item at a time during library operations and direct rating overlays, so processing 1,000 uncached items could consume 1,000 requests.

This change adds a cache-aware MDBList batch lookup that sends up to 100 IDs per request. Library operations, direct MDBList text overlays, and MDBList-backed overlay `value_filter` evaluation now prefetch the relevant items before their per-item processing. Movie libraries prefer TMDb IDs, show libraries prefer TVDb IDs, and items without the primary provider ID use IMDb as a fallback.

Bulk results are retained in a per-run memory cache as well as the existing persistent cache. This allows operations and overlays to share fetched data and keeps batching effective when `settings.cache` is disabled. Season and episode overlay lookups are deduplicated by parent show, and fresh overlay-value cache entries are excluded from prefetching.

The MDBList batch response can omit fields such as `released` and `released_digital`. The response parser now treats those fields as optional and accepts the alternate `tmdb_id` and `imdb_id` field names returned by the batch endpoint.

For 1,000 uncached items, regression coverage confirms that operations and overlays each issue exactly 10 requests rather than 1,000. Operations followed by overlays reuse the in-memory results without duplicate requests.

Validation:

- Full pytest suite: `1282 passed`
- Black, isort, flake8, and commit hooks passed
- Tests cover warm and disabled persistent caches, partial and sparse responses, request chunking, quota exhaustion, movies, shows, IMDb fallback, parent-show deduplication, operations-to-overlays reuse, and MDBList value filters

## Related Issues [optional]

- None

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [ ] Yes
- [ ] No
- [x] Not Applicable

- [x] Yes
- [ ] No

---
<a href="https://app.blacksmith.sh/Kometa-Team/codesmith/Kometa/pr/3509"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789488716&installation_model_id=431096&pr_number=3509&repository=Kometa-Team%2FKometa&return_to=https%3A%2F%2Fgithub.com%2FKometa-Team%2FKometa%2Fpull%2F3509&signature=71e6723050b70c3317d2f27db3703e759eb2e97c6bb48aa216b07d7482be714f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


modules/builder.py
badja-dev, line 1: `_prefetch_mdblist_value_filters()` doesn't check `config.MDBList.limit` before building the candidate list and calling `library.prefetch_mdblist()`. Wasted work once quota is exhausted (`prefetch_mdblist` itself still bails out, so not harmful, just wasteful)

modules/mdblist.py
badja-dev, line 1: If `_request` raises `LimitReached` partway through the batches, `get_items()` has no try/except, so it propagates up with no results returned. Callers don't use the return value so nothing breaks, but worth a comment in the code so a future reader doesn't assume it's a bug

`_response_id()` silently drops any batch response entry that doesn't match a requested ID (just logs a warning, no count/summary). Could make partial data loss harder to notice in large runs

`batch_size` validation (1-100) raises Failed with a generic message, doesn't include the value that was passed

modules/operations.py
badja-dev, line 1: `_prefetch_mdblist()` only falls back to IMDb when the primary id (tmdb/tvdb) is missing. The actual per-item lookup in `mdb_obj()` also falls back to IMDb when the primary id exists but MDBList has no record for it. Those items still cost an individual request later, so the batching win is smaller than advertised for libraries with gappy MDBList coverage

`config.MDBList.limit` is checked both before the provider loop and inside it. Redundant

modules/overlays.py
badja-dev, line 1: Regex match on properties[over_name].name to detect mdb variables is fragile if overlay text formatting/wrapping changes later.
Same IMDb-fallback gap as before applies here

modules/plex.py
badja-dev, line 1: Same `primary-id-missing-only` IMDb fallback gap as `operations.py`, since this is where operations and overlays both delegate to

badja-dev (dismissed):
No correctness bugs found. One real gap: IMDb fallback in prefetch only triggers on a missing primary id, not on a failed primary-id lookup. Rest is nitpicking

modules/mdblist.py
badja-dev, line 1: `cache_run_alias()` only writes to the in-run cache, not the persistent cache. So the IMDb-fallback result for a primary id that has no MDBList record will be re-fetched via IMDb again on every future run rather than being remembered. Minor, but means the fix doesn't fully carry the efficiency gain across runs
Not a blocker, can go without fixing

PR #3514, opened 2026-08-18
More gracefully handle GitHub outages and rate limits

YozoraXCII:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Improves Kometa’s handling of GitHub outages and rate-limit responses.

- Check GitHub Status before version resolution, reporting on any issues and pointing users to the GitHub Status website.
- Handle exhausted request retries without a critical error during version, public Trakt client-ID, and Anime-IDs mapping requests.
- Treat 429 version responses as unavailable instead of displaying the response body as a version.

## Related Issues [optional]

- Related Issue #
- Closes #

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

---
<a href="https://app.blacksmith.sh/Kometa-Team/codesmith/Kometa/pr/3514"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789635507&installation_model_id=431096&pr_number=3514&repository=Kometa-Team%2FKometa&return_to=https%3A%2F%2Fgithub.com%2FKometa-Team%2FKometa%2Fpull%2F3514&signature=0c5c9becddbada8562e051a6f984667bd48b8a50f043be50af652eb636ea2471"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/modules/request.py
+++ b/modules/request.py
@@ -137,7 +137,7 @@ class Requests:
         if response.status_code == 404:
             raise Failed(f"URL Error: No file found at {url}")
         if response.status_code == 429:
-            raise Failed(f"URL Error: Too many requests -  {url}")
+            raise Failed(f"URL Error: Too many requests - {url}")
         if response.status_code >= 400:
             raise Failed(f"URL Error: {response.status_code} on {url}")
         # get_yaml never sets a path, so save() was already a no-op here - read_only=True is a pure speed win (safe loader) plus a defensive guard against future misuse.
@@ -207,6 +207,8 @@ class Requests:
 
     def get_json(self, url, json=None, headers=None, params=None, header=None, language=None):
         response = self.get(url, json=json, headers=headers, params=params, header=header, language=language)
+        if response.status_code == 429:
+            raise Failed(f"URL Error: Too many requests - {url}")
         try:
             return response.json()
         except ValueError:
--- a/modules/request.py
+++ b/modules/request.py
@@ -8,7 +8,7 @@ import requests
 import ruamel.yaml
 from lxml import html
 from requests.exceptions import ConnectionError, RequestException
-from tenacity import retry, stop_after_attempt, wait_exponential
+from tenacity import RetryError, retry, stop_after_attempt, wait_exponential
 
 from modules import timings, util
 from modules.poster import ImageData
@@ -17,6 +17,9 @@ from modules.util import Failed
 logger = util.logger
 
 image_content_types = ["image/png", "image/jpeg", "image/webp"]
+github_status_url = "https://www.githubstatus.com/api/v2/summary.json"
+github_status_components = ["Git Operations", "Webhooks", "API Requests"]
+github_hosts = ["api.github.com", "github.com", "raw.githubusercontent.com"]
 
 # Per-socket-operation timeout for every outbound request; without one a
 # stalled external server hangs the whole run (tenacity only retries on
@@ -86,6 +89,8 @@ class Requests:
         self._branch = None
         self._latest = None
         self._newest = None
+        self._github_is_operational = None
+        self._github_rate_limit_reported = False
         self.session = self.create_session()
         self.cloudscraper = cloudscraper.create_scraper()
         self.global_ssl = verify_ssl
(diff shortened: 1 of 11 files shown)


YozoraXCII:
@badja-dev whenever you get a chance, could you review this - I'm not sure this is the right approach, so open to this being canned and replaced by another PR

modules/request.py
badja-dev, line 1: **`modules/request.py` (`get_yaml`)** - the 429 message dropped the URL, which also affects `config.py:2605`, `imdb.py:1036/1041`, and `meta.py:139/184` - not just the `github.py` `translation_yaml` re-raise this PR is about. Yoz - intentional, or worth keeping the URL for those?

**`modules/request.py` (`get_json`)** - the new unconditional 429 -> `Failed` also changes `mal.py`'s `_request`/`_jikan_request` and `tautulli.py`'s `_request`, not just the three sources named in the description. Looks like a strict improvement but is untested for those two. Yoz - deliberate scope, or just a side effect of it living in a shared method?

badja-dev (changes requested):
Core fix is solid and well-tested - `_version()` no longer treats a 429 body as a version string, `RetryError` is caught wherever retries can exhaust, CI is green, CHANGELOG entry present. Close to mergeable pending the two questions above

modules/request.py
YozoraXCII: The YAML rate-limit message should keep the URL. That’s useful context for any YAML download failure, not just the GitHub translation case, so I’ve kept it and added a test.

The JSON rate-limit handling is deliberately shared. A 429 means the source has asked us to slow down, so treating it as a clear failure is safer than trying to use whatever response it sent. This also affects MyAnimeList and Tautulli

badja-dev (commented):
Round 1 flagged two open questions; both are now resolved by a `head_ref_force_pushed` that also stripped the PR down substantially - and that shrink is the main thing worth Yoz's confirmation before this merges. CI is green on the current tip. Not "no longer needed" - it's still open and active (Yoz explicitly requested re-review on 2026-08-29) - but the scope is now much narrower than the title/description say, and the CHANGELOG entry that existed at round 1 is gone

Resolved from round 1:
- `get_yaml`'s 429 message: confirmed it never actually dropped the URL in the current diff - this round's change is just the double-space typo fix (`"...requests -  {url}"` -> `"...requests - {url}"`). Every `get_yaml` caller (`config.py`, `imdb.py`, `meta.py`) still gets the URL. Round 1's concern was about an earlier version of this diff; non-issue now.
- `get_json`'s unconditional 429 -> `Failed`: now explicitly tested for both callers round 1 named as untested - `test_mal.py::test_rate_limit_failure_propagates_from_json_request` (both `_request` and `_jikan_request`) and `test_small_modules.py`'s Tautulli equivalent. Deliberate and covered

New:
- **Scope shrink vs. description/title** - the PR title and description still promise "Check GitHub Status before version resolution", handling exhausted retries for version/Trakt-client-ID/Anime-IDs requests, and treating 429 version responses as unavailable. None of that is in the current diff (`modules/request.py`'s `_version()` still only catches bare `ConnectionError`, same as nightly) - the force-push left only the `get_yaml`/`get_json` fix behind. There's no second open PR from Yoz carrying the rest. Yoz - is the GitHub-status/retry piece coming back in a follow-up, or did you intentionally cut it down to just this? Worth updating the description either way so it matches what's actually being reviewed.
- **CHANELOG.md** - round 1 noted an entry was present; it's gone from the current diff and there's no matching line in nightly's CHANGELOG either, despite the PR checklist still saying "Yes". Needs re-adding for whatever the final scope ends up being

PR #3520, opened 2026-08-22
feat: Add FlickList connector with list and personal-data builders

badja-dev:
## What type of PR is this?

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [X] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [ ] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

Adds `flicklist` as a new metadata/list source, following the same shape as the existing Trakt/YamTrack connectors. Authenticates with a single pasted API key (no OAuth in this repo - a separate device-code auth helper page lives in Kometa-Utilities).

New builders: `flicklist_list` / `flicklist_list_details` (any FlickList list by id or URL), `flicklist_user_lists` (all of a user's public lists), `flicklist_watchlist`, `flicklist_favorites`, `flicklist_watched`, `flicklist_ratings` (with optional min/max filtering), `flicklist_up_next` and `flicklist_tracked` (both show-only).

This PR is read-only, no writes to FlickList yet. List-sync (`sync_to_flicklist_list`) and the `flicklist_user` mass-rating source are separate, stacked PRs on top of this one (#3521 and a third) so each can be reviewed independently.

## Have you updated the Documentation to reflect changes (if necessary)?

- [X] Yes
- [ ] No
- [ ] Not Applicable

## Have you updated the JSON Schema files (if necessary)?

- [X] Yes
- [ ] No
- [ ] Not Applicable

## Have you updated the CHANGELOG.md?

- [X] Yes
- [ ] No

PR #3521, opened 2026-08-22
feat: Add FlickList list sync (sync_to_flicklist_list)

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Stacked on #3520. Adds `sync_to_flicklist_list` (mirrors `sync_to_trakt_list`): after a collection builds, its resolved movies/shows are pushed to a FlickList list by id or name (auto-created if the name doesn't match an existing list), with items no longer in the collection removed from the list. `sync_missing_to_flicklist_list` optionally also includes movies/shows the collection wanted but couldn't find in Plex.

Matching between Kometa's desired set and FlickList's current list contents is done by comparing every usable id on each side (tmdb, tvdb, imdb, fldb - FlickList's own tvdb-to-tmdb resolution reuses Kometa's existing `Convert` cache, so this never touches FlickList's own request budget). A list item is only removed when it shares no id at all with anything currently wanted, which is the conservative-on-delete behavior the design calls for.

Writes are batched at 1000 items per FlickList API call, with `Retry-After` honored on 429s and capped so a very long rate-limit window fails the run instead of blocking it for up to an hour.

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `url_theme` and `file_theme` metadata attributes for uploading theme music to individual movies and shows.
 - Support the `folder_location` Plex search option in music-library track builders.
 - Add a FlickList connector (`flicklist` config attribute) with read-only `flicklist_list`, `flicklist_list_details`, `flicklist_user_lists`, `flicklist_watchlist`, `flicklist_favorites`, `flicklist_watched`, `flicklist_ratings`, `flicklist_up_next`, and `flicklist_tracked` builders, plus a `flicklist_description` summary source.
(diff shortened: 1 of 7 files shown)


.pyright-baseline.json
```diff
@@ -1,12 +1,13 @@
...
-  "total_errors": 8,
+  "total_errors": 9,
   "per_file_errors": {
     "kometa.py": 1,
     "modules/builder.py": 1,
     "modules/config.py": 1,
     "modules/meta.py": 1,
-    "modules/operations.py": 2,
+    "modules/operations.py": 1,
+    "modules/overlay.py": 2,
```
YozoraXCII: Not sure why the pyright for overlay.py is increased if this PR doesn't touch it

badja-dev: Because of some horrific merging nonsense I had to :D Can be left out

YozoraXCII: Do you want to update the PR or should I just merge?

badja-dev: commit is fine

PR #3553, opened 2026-09-08
chore: pin GitHub Actions to commit SHAs and group Dependabot action updates

badja-dev:
## What type of PR is this?

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ ] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [X] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

Pins all 117 `uses:` references across 18 workflows to full commit SHAs, with the exact version in a trailing comment. Stack 2 of 5, based on #3552.

Tags are mutable. Anyone able to push to an action's repository can move `v7` to different code, which is how `tj-actions/changed-files` was compromised in March 2025 (CVE-2025-30066) - an action this repository uses. A commit SHA cannot be moved.

The two `./.github/workflows/*.yml` reusable workflow calls are intentionally left unpinned.

Two normalisations are included:

- `docker/login-action` was split between `@v4` and `@v4.5.2` across workflows. Both now resolve to v4.6.0.
- `docker/setup-qemu-action@master` now points at v4.3.0, which is the commit `master` currently points at, so it is a no-op in behaviour.

Also adds a `groups` block to `.github/dependabot.yml`. SHA pinning means every action release produces an update, and Dependabot runs daily here. Ungrouped that is one PR per action per release; grouped it is one PR. The pinning does not survive contact with the repository without it.

Dependabot updates SHA-pinned actions and rewrites the trailing version comment, so these stay current automatically.

Follow-up needed in other repositories: `Kometa-Team/discord-notifications` and `Kometa-Team/tag-new-version` have no tags at all, so Dependabot cannot resolve their pinned SHAs back to a release (dependabot-core #14716) and will leave those comments stale. Tagging both fixes it.

No CHANGELOG entry: workflow and CI configuration only.

## Have you updated the Documentation to reflect changes (if necessary)?

- [ ] Yes
- [ ] No
- [X] Not Applicable

## Have you updated the JSON Schema files (if necessary)?

- [ ] Yes
- [ ] No
- [X] Not Applicable

## Have you updated the CHANGELOG.md?

- [ ] Yes
- [X] No

PR #3555, opened 2026-09-08
fix: pin git dependencies to immutable commit SHAs

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

`requirements.txt` installed `serializd-py` from a bare git URL with no ref at all. Stack 3 of 5, based on #3553.

```
serializd-py @ git+https://github.com/Velocidensity/serializd-py.git
```

That resolves to whatever sits at the default branch HEAD of a third-party repository at build time. Two identical builds a day apart can contain different code, nothing is reviewed, and nothing is reproducible. `serializd` is imported at runtime in `modules/builder.py`, so this is in the shipped image.

`docs/requirements.txt` and `[tool.uv.sources]` had a milder form of the same problem, pinned to `@master` on a Kometa-owned fork.

Both are now pinned to commit SHAs:

- `serializd-py` at `1b34bb2ddee12df12bfd19c88d95d489a0bcd483`
- `mkdocs-include-markdown-plugin` at `027bca533839d23fdd7552d68d37d40d8dd40ec4`

This also adds `serializd-py` to `[project].dependencies`, where it was missing despite being a runtime import. It was previously declared only in `requirements.txt` and the dev group, so `uv sync` without `--group dev` produced an environment missing a runtime dependency.

Follow-up: `docs/requirements.txt` now installs from a bare commit SHA, which is immutable but only as durable as that commit. A tag on `Kometa-Team/mkdocs-include-markdown-plugin` pointing at the same commit would be immutable and durable.

No CHANGELOG entry: no user-facing behaviour change.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [X] No


chazlarson (approved):
LGTM

PR #3556, opened 2026-09-08
chore: move documentation dependencies into their own group

badja-dev:
## What type of PR is this?

- [ ] Bug Fix (non-breaking change which fixes an issue)
- [ ] Feature/Tweak (non-breaking change which adds new functionality or enhances existing functionality)
- [ ] Breaking Change (fix or feature that would break any existing functionality for users)
- [ ] Documentation Update
- [X] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

`[project].dependencies` declared 42 packages, 17 of which are the mkdocs documentation toolchain. Stack 4 of 5, based on #3555.

None of those 17 are imported anywhere in `modules/` or `kometa.py`. They exist only to build the documentation site. This moves them into a `docs` dependency group. `docs` is not a default group, so `uv sync` no longer pulls the documentation stack into a runtime environment.

After the split, `[project].dependencies` and `requirements.txt` agree exactly:

```
[project].dependencies: 25    requirements.txt: 25
in pyproject only: none
in requirements only: none
version specifier mismatches: none
```

Two lists that have been maintained by hand in separate files now match package for package and specifier for specifier.

`docs/requirements.txt` is unchanged, so the Read the Docs build is unaffected.

This is a prerequisite for #3557, which generates `requirements.txt` from the lockfile. Without the split, the generated file would pull the entire documentation stack into every image.

No CHANGELOG entry: no user-facing change.

## Have you updated the Documentation to reflect changes (if necessary)?

- [ ] Yes
- [ ] No
- [X] Not Applicable

## Have you updated the JSON Schema files (if necessary)?

- [ ] Yes
- [ ] No
- [X] Not Applicable

## Have you updated the CHANGELOG.md?

- [ ] Yes
- [X] No

PR #3557, opened 2026-09-08
chore: make uv.lock the authoritative install path

badja-dev:
## What type of PR is this?

- [ ] Documentation Update
- [X] Chore (maintenance, dependency bumps, housekeeping - no functional change)
- [ ] Other

## Description

Makes `uv.lock` the file the image is installed from, rather than a second description of the dependencies that has to be kept in step by hand. Stack 5 of 5, based on #3556.

`Dockerfile.base` now runs `uv sync --locked --no-dev --no-install-project` instead of `pip3 install -r requirements.txt`. The lockfile records a sha256 for every registry package (1121 of them) and a commit SHA for each git dependency, so all of that is verified at install time.

`requirements.txt` becomes generated by `uv export`, growing from 25 hand-written direct dependencies to the full 61 package runtime closure with every transitive pinned. It stays an ordinary pip requirements file, so the local install path in the documentation is unaffected.

CI gains a check that fails the PR if either file has drifted:

```
uv lock --check
uv export --format requirements-txt --no-hashes --no-dev --no-emit-project -o requirements.txt
git diff --exit-code -- requirements.txt
```

The uv version is pinned there so CI cannot rewrite the lockfile revision header.

### Why not pip --require-hashes

Hashes in a requirements file are all or nothing, and a VCS dependency can never have one. Both pip and uv refuse:

```
error: In `--require-hashes` mode, all requirements must have a hash, but none
were provided for: mkdocs-include-markdown-plugin @ git+https://...
```

`serializd-py` is not published on PyPI, so it cannot be converted into a hashable registry dependency. Installing from the lockfile is the only route that verifies hashes and still handles a git dependency.

### No user-visible change

`ENV PATH="/.venv/bin:$PATH"` makes `python3` resolve to the virtualenv, so `ENTRYPOINT` is untouched and `docker exec` gets the same environment. Both the currently published base image and the new one work with the same app layer, so the transition is safe in either direction.

Verified locally:

```
docker build -f Dockerfile.base                            succeeds
which python3                                              /.venv/bin/python3
python3 -V                                                 Python 3.13.15
import plexapi, requests, lxml, PIL, tmdbapis, serializd   ok
image size                                                 573MB (published base is 585MB)
```

`uv` is bind-mounted into the build from a digest-pinned stage rather than copied in, so the 50MB binary never becomes a layer in the shipped image.

### Also included

- A `docker-base-build` CI job. `Dockerfile.base` was previously built only by `docker-base.yml`'s weekly schedule, which checks out nightly, so changes to it were never exercised before merge. The new job builds it on any PR touching `Dockerfile.base`, `pyproject.toml` or `uv.lock`, then asserts the runtime imports resolve and that `python3` points at the locked environment. Build only, never pushed.
- `.venv` added to `.dockerignore`. It contained `venv` but not `.venv`, and `.envrc` creates one, so `COPY . /` could paste a developer's host virtualenv over the image's.
- `pyproject.toml` and `uv.lock` added to the docker build trigger paths, since they now change what the image contains.

No CHANGELOG entry: build and CI configuration only, no user-facing change.

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [X] No


chazlarson (approved):
Looks good, and learned a new thing, the "as uv" thing.

PR #3563, opened 2026-09-13
Fix malformed TMDb TVDb mappings

### LINKED ISSUE #3548: [Defaults]: TVDb Error: `tt3348258` fails to map to a TVDb URL when using the `imdb` collection default
### Link to File

n/a

### Version Number

2.4.8

### What branch are you on?

master

### Describe the Bug

hopefully I'm not posting this to the wrong place.

Every run produces the following error, sourced from the built-in `default: imdb` collection file applied to my TV Shows library:

```
TVDb Error: tt3348258 must begin with https://www.thetvdb.com/movies/ or https://www.thetvdb.com/series/
```

`tt3348258` is the IMDb ID for *Arrow: Blood Rush*, a 6-episode online mini-series tied to *Arrow* (2012). It doesn't appear to have a standalone TVDb series/movie page, so the IMDb → TVDb mapping used internally by the `imdb` defaults file fails and logs this error on every run.

The item isn't in my library, this is coming purely from the defaults data set, not anything in my own config.

## Expected Behavior
Either:
- The `imdb` defaults data should skip/exclude IDs with no valid TVDb mapping instead of raising a hard error, or
- The error should be downgraded to a warning/info line (similar to the existing "No TVDb ID found for TMDb IDs" convert-summary lines), since there's nothing a user can fix on their end for an ID that Kometa itself can't map.

## Steps to Reproduce
1. Configure a TV Shows library with `collection_files: [{default: imdb}]`
2. Run Kometa
3. Observe the TVDb Error for `tt3348258` in the log/summary every run

## Logs
```
[ERROR] | TVDb Error: tt3348258 must begin with https://www.thetvdb.com/movies/ or https://www.thetvdb.com/series/ |
```
(Reproduces consistently across multiple runs/days — see `meta-1.log` through `meta-9.log`.)

## Workaround
Adding `tt3348258` to `ignore_imdb_ids` in `config.yml` suppresses the error, but it would be good if the defaults file didn't require every user to individually discover and ignore this ID.

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

TMDb can return a malformed TVDb external ID for a show. In the reported case, the TMDb record used `tt3348258`—an IMDb ID—in its TVDb ID field. `Convert.tmdb_to_tvdb()` previously accepted any truthy value, so Kometa returned and cached that string as a TVDb ID. The IMDb Defaults builder then added it to `missing_shows`, where the TVDb client interpreted it as a URL and logged a misleading hard error on every run.

This PR validates TVDb IDs at the TMDb-to-TVDb conversion boundary and normalizes valid numeric strings to integers.

| Input/source | Previous behavior | New behavior |
| --- | --- | --- |
| Numeric cached TVDb ID | Returned | Returned as an integer |
| Numeric string from TMDb | Returned and cached as supplied | Normalized to an integer, then cached |
| Malformed cached value | Returned as a TVDb ID | Ignored; TMDb is queried again in case upstream data was corrected |
| Malformed value from TMDb, such as `tt3348258` | Returned/cached and later sent to the TVDb URL parser | Not cached; treated as an ordinary missing conversion |

For the IMDb show builder, the ordinary missing-conversion behavior is already to emit a `Convert Warning` and skip the item. As a result, the invalid value no longer enters `missing_shows` and never reaches `TVDb.get_tvdb_obj()`.

### Scope

- No special case for `tt3348258` or the IMDb Defaults file.
- No changes to valid numeric TVDb mappings.
- No database migration or cache deletion.
- No changes to TVDb request, retry, or NotFound behavior.
- Existing malformed cached values are bypassed and can be overwritten if TMDb later supplies a valid ID.

### Regression coverage

| Test area | Coverage |
| --- | --- |
| Conversion | Numeric strings normalize to integers |
| Conversion | Malformed live TMDb external IDs return a conversion miss and are not cached |
| Cache | Malformed cached mappings are rechecked against TMDb |
| Error contract | `fail=True` retains the existing `MappingConvertError` / `Convert Warning` behavior |
| IMDb builder | The affected item is skipped, is not added to `missing_shows`, and does not produce a normal error |

### Verification

- `pytest -q`: **1544 passed**
- Focused suite (`tests/test_convert.py tests/test_builder.py`): **143 passed**
- Targeted regression set: **11 passed**
- `black`: passed
- `isort`: passed
- `flake8`: passed
- `git diff --check`: passed

The changelog documents the user-visible correction. Additional user documentation is not needed because this does not add or change configuration, builders, attributes, or supported values.

## Related Issues [optional]

- Closes #3548

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No

---
<a href="https://app.blacksmith.sh/Kometa-Team/codesmith/Kometa/pr/3563"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791850544&installation_model_id=431096&pr_number=3563&repository=Kometa-Team%2FKometa&return_to=https%3A%2F%2Fgithub.com%2FKometa-Team%2FKometa%2Fpull%2F3563&signature=f6de4186a607a07b1acdb8946c1b5f8bbfa4fcf20f4b992424391c447f98a30e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


modules/convert.py
badja-dev, line 1: `normalize_tvdb_id` uses `.isdigit()`, which is `True` for characters `int()` will not accept - `"²".isdigit()` is `True` but `int("²")` raises `ValueError`, so the guard can still let an unhandled exception through where it is meant to return `None`; `.isdecimal()` is the exact test for "`int()` will accept this" and is a one-word change

@antwanchild  - on the malformed-cached-value path, the row is bypassed but never cleared, so while TMDb keeps returning the bad external ID that show costs an extra TMDb call on every run forever - is it worth keeping the door open for upstream correcting the data, or worth clearing the row?

antwanchild: Good catch — changed this to `.isdecimal()` in 674da6d6 and added a regression test using `"²"`.

**Cache behavior (no code change):** I intentionally did not clear the malformed cached row. Clearing it would not reduce TMDb calls: it would turn the malformed-cache path into an ordinary cache miss, and both paths query TMDb again. Keeping the legacy row also does not block recovery—the method bypasses the malformed value, queries TMDb, and overwrites the row as soon as TMDb returns a valid TVDb ID. New malformed responses are never cached.

Actually suppressing repeated requests would require negative-cache expiry semantics. That would also delay detecting an upstream correction until the negative entry expires, so it is a broader behavior change than this focused crash fix.

PR #3567, opened 2026-09-13
Add Tracearr watched media builder

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Adds `tracearr_watched_media`, an opt-in builder backed by Tracearr's compact Public API v2 `/watched-media` endpoint.

This remains separate from `tracearr_watched` because the two builders answer different questions:

| Builder | Result | Time behavior | Ordering |
|:--|:--|:--|:--|
| `tracearr_watched` | Titles ranked by completed playback sessions | Uses a required history window (`list_days`, default 30) | Completed-session count |
| `tracearr_watched_media` | Distinct watched media, or watched plus partially watched media | All-time by default; optional `list_days` filter | Most recent watched day |

The new builder supports movie, show, and episode-level results; `user`, `min_state`, `list_days`, and `list_size`; cursor pagination; mixed movie/show playlists; and provider-ID-first matching.

| Media type | Preferred matching |
|:--|:--|
| Movie | TMDb, then IMDb |
| Show | TVDb, then TMDb or IMDb |
| Episode | Parent-show TVDb ID + season/episode coordinates, then episode IMDb |

Kometa now retains the v2 OpenAPI path catalog discovered during connection and only enables this builder when Tracearr advertises `/api/v2/public/watched-media`. Older Tracearr installations are unaffected, and using the builder against an unsupported version produces a focused compatibility error rather than silently changing its semantics through a history fallback.

The endpoint allows 1,000 compact records per page versus 100 full playback records per history page:

| API characteristic | `/history` | `/watched-media` | Integration benefit |
|:--|:--:|:--:|:--|
| Maximum page size | 100 plays | 1,000 media records | Up to 10x fewer page requests for equivalent row counts |
| Result grain | One record per play | One record per distinct title | No local play aggregation for watched-state sets |
| Identity data | Full playback and media identity | Compact provider-focused identity | Less transferred data for library matching |

Existing history builders also pass an explicitly configured `watched` filter to API v2 so Tracearr can discard unwanted records before transfer. The in-progress builder deliberately keeps its existing behavior because it needs both partial and completed plays to determine the latest state correctly.

Validation:

| Check | Result |
|:--|:--|
| Full pytest suite | 1,548 passed |
| Focused Tracearr/builder/schema suite | 234 passed |
| Black | Passed |
| isort | Passed |
| flake8 | Passed |
| Pyright baseline | Passed; 0 regressions |
| pyspelling | Not run locally because `aspell` is unavailable |

The Tracearr defaults are intentionally unchanged. The new endpoint is not yet universally available, and the existing default `Tracearr Watched` chart has different semantics.

## Related Issues [optional]

- Related Issue # N/A
- Closes # N/A

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No
- [ ] Not Applicable

- [X] Yes
- [ ] No

---
<a href="https://app.blacksmith.sh/Kometa-Team/codesmith/Kometa/pr/3567"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791856667&installation_model_id=431096&pr_number=3567&repository=Kometa-Team%2FKometa&return_to=https%3A%2F%2Fgithub.com%2FKometa-Team%2FKometa%2Fpull%2F3567&signature=fca584a20ad8165e803d54942ff91659ab9878ef9b860a8a23c098019336a847"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


docs/files/builders/tracearr/history.md
badja-dev, line 1: The new `## Tracearr Watched Media` section is inserted at line 50, immediately above the `tracearr_binged` (line 51) and `tracearr_in_progress` (line 53) paragraphs, and this file has no other `##` heading anywhere - it is `# Tracearr History` at line 5 and then `### Example Tracearr History Builder(s)` at line 59. The result is that everything from line 51 to the end of the file is absorbed into the Watched Media section, so the two paragraphs about other builders, the configuration warning, and the general history examples all render as children of Watched Media in the right-hand navigation. Ant - either moving the new section below the examples at the bottom of the page, or adding a `## Tracearr History Builders` heading above line 51 to close it off, fixes it; the `#tracearr-watched-media` anchor that `builders.json` and both overview tables point at stays valid either way. The content itself is good and the attribute table correctly marks all four attributes optional.

modules/builder.py
badja-dev, line 1: The early return for `watched_media` keeps it out of the shared history `final_dict`, which is the right call - the history path would otherwise have forced `list_days` and the progress attributes onto a builder that does not want them - and leaving `list_days` and `user` as `None` when absent rather than defaulting them is what makes the all-time default actually work.

modules/tracearr.py
badja-dev, line 1: The new `watched` passthrough in `get_rating_keys` reads `data.get("watched") is not None`, which really does mean "explicitly configured" since `_tracearr` sets it to `None` for unconfigured history builders (`builder.py:3375`), and excluding `list_type != "in_progress"` correctly stops `in_progress`'s forced `watched = False` (`builder.py:3387`) from being sent as an API filter; the local filter at line 297 is untouched, so this stays a pure transfer optimisation that cannot change results even if Tracearr's server-side semantics differ from Kometa's. The provider-ID matching also lines up with the downstream resolver on all four types - `tvdb_episode` emits exactly the `{tvdb}_{season}_{episode}` string that `builder.py:4036` splits, `tmdb_show` is handled at `builder.py:4051`, and the episode IMDb fallback resolves through the `imdb_to_tmdb` episode branch at `builder.py:3924` - and sorting on `candidate[0]` instead of the whole tuple is worth calling out as correct, since mixed `int` and `str` IDs would raise on a plain reverse sort. Two things worth your word on: `exhausted_by_cutoff` stops pagination at the first record older than the cutoff, which is only correct if `/watched-media` guarantees `last_watched_day` descending order - the docs assert that but the code takes it on trust, so one out-of-order record silently under-fills the list - and the builder calls `_request` directly rather than going through `_fetch_history`, so it gets none of the per-run request reuse the history builders get and two identical `tracearr_watched_media` definitions in one run will each re-paginate from scratch. @antwanchild - is that ordering contractual on Tracearr's side, and was bypassing the run cache deliberate? Two nits if you are back in this block anyway: `getattr(self, "v2_paths", set())` is defensive for an attribute `__init__` always sets, and an empty `media_types` returns an empty list rather than raising the way the other unsupported-configuration paths do.

badja-dev (commented):
The new `## Tracearr Watched Media` heading in `docs/files/builders/tracearr/history.md` is the first `##` in a file that previously went straight from H1 to a single H3, so it swallows everything after it: the `tracearr_binged` and `tracearr_in_progress` paragraphs, the Tracearr Configuration warning, and the `### Example Tracearr History Builder(s)` block with the generic history YAML all end up nested under Watched Media in the page navigation

docs/files/builders/tracearr/history.md
antwanchild: Addressed in 13b46fb2 by adding a `## Tracearr History Builders` heading before the general history content. This closes the Watched Media section while preserving the existing `#tracearr-watched-media` anchor.

modules/builder.py
antwanchild: Thanks for confirming this. I kept the early-return parser behavior and the `None` defaults unchanged, so the all-time default and separation from history-only attributes remain intact. No code change was needed for this thread.

modules/tracearr.py
antwanchild: Addressed all four points in 13b46fb2:

- Ordering is contractual: Tracearr’s OpenAPI description says the endpoint returns newest activity first, and the server query orders by `last_day DESC` with canonical ID as its tiebreak. I retained the cutoff optimization and added a comment documenting that invariant.
- Bypassing per-run reuse was not necessary. Watched-media pages now have a dedicated per-run cache keyed by their request parameters, so identical definitions reuse pagination and differently sized definitions can reuse the pages they share. A regression test verifies this.
- Removed the defensive `getattr`; `self.v2_paths` is used directly.
- An empty supported-media scope now raises a focused `Failed` error rather than returning an empty list.

The positive notes about server-side `watched` filtering, provider matching, and mixed-ID sorting remain unchanged.

PR #3571, opened 2026-09-13
Fix Plex language search normalization and diagnostics

### LINKED ISSUE #3570: Fix Plex language search normalization and diagnostics
## Version Number

2.4.8

## What branch are you on?

master

## Describe the Bug

Plex language searches regress in three related ways after language values began expanding to every Plex locale/code variant:

1. Language names such as `Dutch` and ISO 639-3 codes such as `nld` are rejected even when Plex reports matching Dutch streams. `get_language_search_values` normalizes Plex choice keys but looks up the original configured value instead of its normalized base code, and it does not consider Plex choice titles.
2. In 2.4.8, a base code such as `nl` expands into multiple positive filters joined with `AND`, which can produce `Plex Error: No Items found in Plex`. Current nightly already groups positive variants with `OR`, but the remaining normalization problems persist.
3. Validation reports `Options: []` because the language-specific path bypasses `get_search_choices` and initializes the available-name list as empty. Expanded variants are also rendered as duplicate filter-description lines instead of one configured-language line.

The negative expansion itself must remain an `AND` across variants so every English variant is excluded.

Reported in Discord: https://discord.com/channels/822460010649878528/1548764007089971220

Related support report: https://discord.com/channels/822460010649878528/1539762336137609226/1539762336137609226

### Expected behavior

- `Dutch`, `nld`, and `nl` resolve to the Dutch values Plex exposes for the library.
- An explicitly available locale/code remains targetable exactly.
- Positive variants are ORed and negative variants are ANDed.
- Invalid values show the available Plex language options.
- The filter summary prints the configured language once.

## Relevant Collection/Overlay/Playlist Definition

```yml
NL:
  plex_search:
    all:
      audio_language: Dutch
      audio_language.not: English
      label.not: Vlaams
  overlay:
    name: NL
    file: config\\overlays\\NL.png
    horizontal_offset: 25
    horizontal_align: right
    vertical_offset: 25
    vertical_align: top
```

## Logs

```text
Plex Error: audio_language: Dutch not found
Options: []
```

Using `nl` validates, but the generated filter reports repeated language lines and returns no items on 2.4.8.

antwanchild:
## What type of PR is this?

- [ ] Documentation Update
- [ ] Other

## Description

Completes the nightly Plex language-filter fix for `audio_language` and `subtitle_language`. Nightly correctly groups multiple locale variants with `OR` for positive searches and `AND` for exclusions, but longstanding configurations using Plex language names such as `Dutch` and `English` still fail validation. Three-letter codes such as `nld` can also fail when Plex does not expose that exact key.

This change indexes Plex language-choice titles alongside exact and normalized codes, falls back from an unavailable three-letter code to the matching base-language variants, reuses cached language names in validation errors, and renders a configured language only once in the filter summary. Name matching uses the choices already returned and cached for that Plex library, so it supports other languages and localized Plex titles without a hard-coded name table or an additional Plex request.

| Input | Current nightly | With this PR |
| --- | --- | --- |
| `Dutch` / `English` | Rejected | Matches Plex-reported language names |
| `nld` | Rejected unless Plex exposes that exact key | Falls back to available Dutch variants |
| `nl` | Matches all Dutch variants | Unchanged |
| Invalid value | `Options: []` | Lists Plex-reported languages |
| Multiple variants | Repeated summary lines | One summary line; query semantics unchanged |

Explicit locale-tagged and three-letter values exposed by Plex remain exact matches. Bare two-letter base codes expand to every available variant of that language. Positive variants remain grouped with `OR`, while negated variants remain grouped with `AND` so every excluded-language variant is rejected.

### Verification

- Full suite: `pytest -q tests` — **1,563 passed**
- Focused language and search tests — **34 passed**, including the reported `Dutch` plus `English` exclusion configuration
- `black` — passed
- `isort` — passed
- `flake8` — passed
- `git diff --check` — passed

## Related Issues [optional]

- Related Issue #3440
- Closes #3570

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [ ] Yes
- [ ] No
- [X] Not Applicable

- [X] Yes
- [ ] No


modules/builder.py
badja-dev, line 1: Since `sub_display` now takes whichever variant is first in the list `get_language_search_values` returns, and that order comes from Plex's own tag-report order rather than the value the user actually typed, the summary could show e.g. "de-DE" instead of "de" depending on how Plex orders its choices for that library - @antwanchild , is that acceptable, or should it prefer the variant that matches what was configured when one exists?

modules/plex.py
badja-dev, line 1: The new `if choice.key not in code_map.setdefault(...)` / `name_map.setdefault(...)` dedup guards - was that guarding against a real duplicate-choice case you hit from Plex, or just defensive? Doesn't affect correctness either way

badja-dev (commented):
Just a couple of questions, but mergeable either way I think

modules/builder.py
antwanchild: It already prefers the configured value for the summary: each variant calls `build_url_arg(..., arg_s=og_value)`, so `sub_display` captures the original configured spelling/case rather than the first Plex variant. I made that behavior explicit in b754be4a by returning `de-DE` before `de` and asserting that the summary still shows only `Audio Language is de`.

modules/plex.py
antwanchild: These guards are defensive; I did not observe duplicate choices in the reported Plex response. I kept them because identical keys from Plex would otherwise produce redundant query clauses, while preserving first-seen order costs very little. Multiple distinct keys with the same title are still retained intentionally in `name_map`.

